PR #20, opened 2020-08-18
Make the api package a dedicated module

### LINKED ISSUE #40: Make api/ its own module
(A recent example of doing this is https://github.com/fluxcd/image-reflector-controller/pull/27)

hiddeco:
This includes changes to:

* Bump the Kubernetes dependencies to v1.19.3 to align
  with the other toolkit controllers.
* Update controller-runtime dependency to v0.6.3 to align
  with the other toolkit controllers.
* Update the source-controller dependency to the most recent
  v0.2.2 version to include the v1beta1 API.
* Add the `pkg/gittestserver` dependency for the Git tests,
  as this package was removed in a newer source-controller version.
* Bump the Go version to v1.15.

Fixes #40 
Supersedes #19


go.mod
```diff
@@ -1,25 +1,27 @@
 module github.com/fluxcd/image-automation-controller
 
-go 1.14
+go 1.15
```
squaremo, line 3: Why is go 1.15 required?

squaremo (approved):
Thanks for this follow-up Hidde 🍇

go.mod
squaremo: A: it is not specifically required; however, since it's difficult to avoid providing _some_ value here, and in the absence of any version-specific features being needed, it's better to make it match the version used in e.g., CI.

Issue #12, opened 2020-08-05
Adapt update to other kinds of workload object


squaremo:
The path to find image references is hard-wired to `.spec.template.spec.{containers,initContainers}[*].image` (in effect; YAML path is not actually used). At least CronJobs have a different structure, so will fail to be updated.

The simplest thing to accomodate CronJob would be to dispatch on the `Kind`, using a different path-traversing func.

To update `HelmRelease` will need a deeper change; I'll have to look at the code in flux and see if I can adapt that to kyaml.

squaremo:
CronJobs now supported via #21.

squaremo:
Closing this, as the setters implementation (which can be used with any workload, or indeed any YAML that has `apiVersion` and `kind` fields) is now the only implementation used. The simple update mechanism was removed in #27.

PR #21, opened 2020-08-27
Apply image updates to CronJob objects

bigkevmcd:
This updates the image replacement to switch on the kind of the node,
and change the way the replacements are done for CronJobs.

CronJobs have PodTemplateSpecs embedded deeper in the structure.

This addresses #12

### DIFF
--- a/pkg/update/update.go
+++ b/pkg/update/update.go
@@ -56,21 +56,35 @@ func makeUpdateImagesFilter(originalRepo, replacement string) kio.Filter {
 	})
 
 	return kio.FilterFunc(func(objs []*yaml.RNode) ([]*yaml.RNode, error) {
+		tees := []yaml.Filter{
+			yaml.Tee(
+				yaml.Lookup("initContainers"),
+				replaceImageInEachContainer,
+			),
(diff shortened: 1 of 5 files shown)


pkg/update/update.go
```diff
@@ -56,21 +56,37 @@ func makeUpdateImagesFilter(originalRepo, replacement string) kio.Filter {
...
-					yaml.Lookup("initContainers"),
-					replaceImageInEachContainer,
-				),
-				yaml.Tee(
-					yaml.Lookup("containers"),
-					replaceImageInEachContainer,
-				),
-			); err != nil {
-				return nil, err
+			switch kind(obj) {
```
squaremo, line 71: As the case bodies differ only on the path given to `yaml.Lookup`, could just that be varied inside the switch and the pipe be run with that value afterward? Or are you anticipating other varieties of pipe.

squaremo (approved):
Jolly good, thank you 🍏

pkg/update/update.go
bigkevmcd: Good point, I've simplified this a bit.

PR #23, opened 2020-10-06
Implement automation via kyaml setters2

squaremo:
This implements the design given in https://github.com/fluxcd/toolkit/discussions/107#discussioncomment-82746; briefly, you mark fields in your YAMLs with an image policy name, and the automation sets those fields (wherever they are) to the latest image for the given policy.

This works but needs some polish. It's not obvious, for example, why something doesn't get updated when you think it should. I also need to try it with a few example configs; e.g., with a kustomization, and with a custom resource.

One choice that was forced upon me: I want the markers to have both a namespace and name of the corresponding image policy, to avoid collisions, but `/` is unavailable as a separator, since it's used in OpenAPI ref paths (and that's what the marker is, a ref). I've used `:`, but any character that's not allowed in resource names and isn't `/` either would do.

I've adapted the instructions [given in the README](https://github.com/fluxcd/image-automation-controller/blob/use-kyaml-setters2/README.md) for trying this out.

TODO (not necessarily in this PR):

 - [ ] respect the `paths` field from the spec
 - [ ] better troubleshooting -- it's hard to work out why it's not working when it's not working
 - [x] ability to replace the tag value rather than the whole image ref (per Stefan's question)
 - [ ] tool to add marker? or at least, the lib for integration into `gotk`
 - [ ] update to v1beta1 (or do that in another PR and rebase here)


stefanprodan:
@squaremo I can't tell from code, can we use a setter to update tags that are not in the same node as the image? e.g.

```yaml
apiVersion: kustomize.config.k8s.io/v1beta1
kind: Kustomization
resources:
  - deployment.yaml
images:
  - name: fluxcd/kustomize-controller
    newName: fluxcd/kustomize-controller
    newTag: v0.1.0 # {"$imagepolicy": "gotk-system:controll-plane"}
```

squaremo:
> @squaremo I can't tell from code, can we use a setter to update tags that are not in the same node as the image? e.g.
> 
> ```yaml
> apiVersion: kustomize.config.k8s.io/v1beta1
> kind: Kustomization
> resources:
>   - deployment.yaml
> images:
>   - name: fluxcd/kustomize-controller
>     newName: fluxcd/kustomize-controller
>     newTag: v0.1.0 # {"$imagepolicy": "gotk-system:control-plane"}
> ```

That's somethin in the design I've not implemented yet -- separating out the bits of the latest image, so you can choose the bit you want. It would look something like this:

```yaml
apiVersion: kustomize.config.k8s.io/v1beta1
kind: Kustomization
resources:
  - deployment.yaml
images:
  - name: fluxcd/kustomize-controller
    newName: fluxcd/kustomize-controller
    newTag: v0.1.0 # {"$imagepolicy": "gotk-system:control-plane:tag"}
```

(NB the `:tag` appended to the marker ref; arbitrarily, I've just used a colon as the separator again)

stefanprodan:
> (NB the :tag appended to the marker ref; arbitrarily, I've just used a colon as the separator again)

Awesome 💯

stefanprodan (approved):
LGTM 🚀

squaremo:
Merging, and making issues for those TODOs ...

PR #27, opened 2020-10-22
Remove proof-of-concept update mode

squaremo:
The initial implementation of image updates required a single image policy, and updated every image field that used the image, in the git repo. This mode has limited practical value, and rather than elaborating on it, it would be better to concentrate on making the more carefully thought-through "setters" mode.


stefanprodan (approved):
LGTM

PR #37, opened 2020-10-30
Sync to latest

### LINKED ISSUE #26: Update to toolkit v1beta1
(not our own apiVersion, but any types used from e.g., source.toolkit.fluxcd.io)

relu:
Fixes #26
Includes #20


go.mod
```diff
@@ -1,23 +1,30 @@
...
+replace github.com/fluxcd/image-automation-controller/api => ./api
 
 require (
-	github.com/fluxcd/image-reflector-controller v0.0.0-20200810165546-c2265d9b49b9
+	github.com/fluxcd/image-automation-controller/api v0.0.0-00010101000000-000000000000
+	github.com/fluxcd/image-reflector-controller/api v0.0.0-20201028160737-240e94730afa
+	github.com/fluxcd/pkg/gittestserver v0.0.2
 	// If you bump this, change TOOLKIT_VERSION in the Makefile to match
-	github.com/fluxcd/source-controller v0.0.7
+	github.com/fluxcd/source-controller v0.2.1
```
squaremo, line 12: I don't think this should be here -- this component should only depend on the API

squaremo: Nope, I'm wrong, it uses source-controller/pkg/git. OK -- ignore me!

relu: Haha, was about to answer you but seems you were quicker ;)

On second thought though, I was actually thinking about how that might actually need to be extracted at some point (considering it's now used in two distinct components). Maybe put under pkg?

relu:
Closing this as it's no longer relevant.

Issue #33, opened 2020-10-26
Multi-arch build and release


squaremo:
See https://github.com/fluxcd/image-reflector-controller/pull/33/files for an example.

squaremo:
Fixed in #42.

PR #42, opened 2020-11-18
Adapt release workflow from image-reflector

squaremo:
I have factored out the controller name into an env entry. In a `run:`
field, it is possible (necessary?) to use shell syntax for referring
to the env entry.

If this works here, I'll backport it to image-reflector-controller.


hiddeco (approved):

squaremo:
Fixes #33.

PR #53, opened 2020-11-24
Make controller update ready condition

### LINKED ISSUE #48: Get conditions and status updates in line
The controller should at least use a Ready condition, even if its interpretation will be a little interesting. (Perhaps check what CronJob resources have for conditions).

Similarly, it probably needs more careful status updates. Look at recording the revision last pushed, that sort of thing.

squaremo:
This gives `ImageUpdateAutomation` objects `.status.conditions` and
`.status.observedGeneration` fields, which are maintained by the
controller in the GOTK-standard way.

The only condition used is a Ready condition, compatible with kstatus
(and in common with other GOTK controllers). An object is marked Ready
if the reconciliation exits without an error, whether or not changes
were actually made. If the automation run cannot proceed, e.g.,
because the git repository referred to does not exist, or is not
cloneable, it will be marked as not ready.

This means the condition is a reliable guide to whether the particular
automation is operating or not; new objects will be marked ready as
soon as they have been run through successfully, and will stay ready
until there's a problem. Generally, if there _is_ a problem, the
object will be requeued with a backoff, or left to wait until
circumstances change (e.g., the object itself is edited); one way or
another, there will be a retry, and thereby an opportunity to
transition to ready.

Fixes #48.


api/v1alpha1/imageupdateautomation_types.go
```diff
@@ -92,6 +94,21 @@ type ImageUpdateAutomationStatus struct {
 	// made).
 	// +optional
 	LastAutomationRunTime *metav1.Time `json:"lastAutomationRunTime,omitempty"`
+	// +optional
+	ObservedGeneration int64 `json:"observedGeneration,omitempty"`
+	// +optional
+	Conditions []metav1.Condition `json:"conditions,omitempty"`
+}
+
+const (
```
relu, line 103: In the other component controllers, these are defined in a separate `condition_types.go` file, I suppose it's not necessarily an officially accepted convention but rather just a relic from the past.

It would be good to have these documented, a short description of what they represent.

controllers/imageupdateautomation_controller.go
```diff
@@ -78,22 +80,41 @@ type ImageUpdateAutomationReconciler struct {
...
 	ctx := context.Background()
 	log := r.Log.WithValues("imageupdateautomation", req.NamespacedName)
+	now := time.Now()
 
 	var auto imagev1.ImageUpdateAutomation
 	if err := r.Get(ctx, req.NamespacedName, &auto); err != nil {
 		return ctrl.Result{}, client.IgnoreNotFound(err)
 	}
 
+	auto.Status.LastAutomationRunTime = &metav1.Time{Time: now}
```
relu, line 90: This would update on every reconcile, regardless of success or failure, is that the intention?

controllers/imageupdateautomation_controller.go
```diff
@@ -78,22 +80,41 @@ type ImageUpdateAutomationReconciler struct {
...
+	if err := r.Status().Update(ctx, &auto); err != nil {
+		return ctrl.Result{}, err
+	}
+
+	// failWithError is a helper for bailing on the reconciliation.
+	failWithError := func(err error) (ctrl.Result, error) {
```
relu, line 96: Maybe this should be extracted to a reconciler method, which would make for a shorter Reconcile func body.

relu (changes requested):

controllers/imageupdateautomation_controller.go
squaremo: I considered that, but thought the convenience of closing over a few of the local variables was worth the extra assignment here.

controllers/imageupdateautomation_controller.go
squaremo: To be honest I did not think hard about exactly when this should be updated. Thinking about what is the purpose of the field (to inform the user), it should be updated when the automation run has completed without error, whether or not it made changes. I'll change it.

api/v1alpha1/imageupdateautomation_types.go
squaremo: I gave each a golint-compatible comment.

controllers/imageupdateautomation_controller.go
squaremo: Move it to the end, just before it sets the ready condition and updates the status for the last time.

relu (approved):
Looking good! :100:

stefanprodan (approved):

PR #55, opened 2020-11-26
Suspend and force mechanisms

### LINKED ISSUE #16: Suspend / Force an automation
The general pattern within GitOps Toolkit is that resources can be suspended (don't act on it), and those running on a schedule can be run on demand (act on it now). See, for example, https://toolkit.fluxcd.io/components/kustomize/kustomization/#reconciliation

squaremo:
This adds the conventional mechanisms for suspending an automation and forcing an automation to run.
(This depends on #53, will rebase and undraft once that's merged)

Fixes #16.


controllers/imageupdateautomation_controller.go
```diff
@@ -78,22 +80,65 @@ type ImageUpdateAutomationReconciler struct {
...
+	// whatever else happens, we've now "seen" the reconcile
+	// annotation if it's there
+	if token, ok := meta.ReconcileAnnotationValue(auto.GetAnnotations()); ok {
+		auto.Status.SetLastHandledReconcileRequest(token)
+		if err := r.Status().Update(ctx, &auto); err != nil {
+			return ctrl.Result{Requeue: true}, err
+		}
+	}
+
+	if auto.Spec.Suspend {
```
stefanprodan, line 99: We decided to not change the Ready status on suspend, instead just do https://github.com/fluxcd/kustomize-controller/blob/main/controllers/kustomization_controller.go#L167-L171

squaremo: Doesn't that break this: https://github.com/fluxcd/flux2/blob/main/cmd/flux/reconcile_kustomization.go#L107 ?

squaremo: If the idea is that it will time out and therefore (eventually) tell the user that the object wasn't reconciled, I would consider reporting that this is probably going happen up front when an object is suspended.

squaremo:
Currently failing because the test for suspend checks that the status is set to unready -- which, after 63dc7e3, is no longer true, since suspending an object effectively prevents anything changing. It is difficult to have a test for nothing happening :-/

controllers/imageupdateautomation_controller.go
stefanprodan: Right, we need lookup the suspended status and tell the user that it should resume the reconciliation first.

squaremo: There's two choices for what happens when someone requests a reconciliation of a suspended resource:

 - time out (probably ... maybe someone else will unsuspend it while you're waiting)
 - trivially succeed

In both these cases it's worth telling the user up front that the resource is currently suspended. But elsewise, the difference is this: do you want the user to have to retry straight away; or is it worth waiting to see if something else changes the resource (they can always interrupt the program if they get bored waiting).

If reconciliation succeeds trivially, you at least know the machinery is running. Whereas if it times out, it could be because the controller is jammed somewhere. I don't think there's much utility in waiting to see if someone else unsuspends an object. For those reasons I prefer succeeding trivially, rather than timing out.

stefanprodan: Fixed in https://github.com/fluxcd/flux2/pull/525

squaremo: I should have had a third option, "bail out without trying". From the user's point of view, this is very similar to trivially succeeding, so :+1:

It doesn't make it any easier to test suspended resources, though -- for that, I recommend updating the reconcile request status before exiting Reconcile, so you can (programmatically) see that the loop has run.

squaremo: > I recommend updating the reconcile request status before exiting Reconcile

.. However in the absence of a convention, I've made my test run Reconcile explicitly.

stefanprodan (approved):
LGTM

PR #56, opened 2020-11-27
Note maintainers shared from fluxcd/flux2

squaremo:
The GitOps toolkit controllers usually adopt the maintainers of the
"central" Flux v2 git repo, fluxcd/flux2. For convenience of keeping
track, it is easier to refer to the MAINTAINERS file there.

NB this expands the maintainers of this repo, but only to those who
were effectively considered maintainers anyway (i.e., it reflects the
established practice.)

See https://github.com/fluxcd/flux2/discussions/515 for the proposal.

### DIFF
--- a/MAINTAINERS
+++ b/MAINTAINERS
@@ -2,7 +2,11 @@ The maintainers are generally available in Slack at
 https://cloud-native.slack.com in #flux (https://cloud-native.slack.com/messages/CLAJ40HV3)
 (obtain an invitation at https://slack.cncf.io/).
 
+In additional to those listed below, this project shares maintainers
+from the main Flux v2 git repository, as listed in
+
+    https://github.com/fluxcd/flux2/blob/main/MAINTAINERS
+
 In alphabetical order:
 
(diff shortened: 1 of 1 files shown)


stefanprodan:
I think we could remove all maintainers and just reference the flux2 file.

hiddeco:
> I think we could remove all maintainers and just reference the flux2 file.

That would not work for controller projects that have maintainers that have a sole interest in just maintaining that project.

I also think there is a difference in being mentioned in the `fluxcd/flux2` maintainers list or in the list of a domain specific project, the latter implies a closer relationship and better expertise.

squaremo:
> > I think we could remove all maintainers and just reference the flux2 file.
> 
> That would not work for controller projects that have maintainers that have a sole interest [...]

For example, I'm not a fluxcd/flux2 maintainer, so that would remove me as a maintainer here :-/

stefanprodan (approved):

hiddeco (approved):

squaremo:
@relu Since you are also a flux2 maintainer about to (implicitly) become a maintainer here, we'll need your stamp too.

relu (approved):
I think this makes sense.

PR #59, opened 2020-11-30
Control if commit message has [skip ci], if yes tests are skipped. Will close #52

### LINKED ISSUE #52: Include [skip ci] in the default commit message
Some people like to keep their configuration alongside their application code, and set up CI to build images from the application code. This can cause problems when combined with Flux's automation, because commits from the automation will cause the CI to run (which builds another image, which causes the automation to run, ...).

Happily, many (most?) CI systems have a "get out clause": if you put a special token in a commit message, they will not run for that commit. For example, Circle CI uses `[ci skip]`: https://circleci.com/docs/2.0/skip-build/ (and someone summarised the support in a whole lot of systems here: https://github.community/t/github-actions-does-not-respect-skip-ci/17325/3).

By including this token in the default commit message (which people can change if they _don't_ want to skip CI for automated commits), the looping problem above will be avoided in a default install.

EDIT: expanded description.

### LINKED ISSUE #54: API user reference for image.toolkit.fluxcd.io/v1alpha1
In addition to docs generated from the API types, by convention there's a more user-oriented reference to the API types. See for example https://toolkit.fluxcd.io/components/helm/helmreleases/ (as opposed to the generated doc, https://github.com/fluxcd/source-controller/blob/main/docs/spec/v1beta1/helmrepositories.md.

Include the image-reflector types as well. This all goes in https://github.com/fluxcd/flux2/tree/main/docs/guides/

fbrv:


squaremo:
First-time contributor high-five! 🙋 

I think you are meaning to address issue #52. That issue is about _the controller_ making commits that will skip users' CI systems, rather than our own CI (GitHub Actions) skipping commits.

To address issue #52, you need to make the _default_ commit message include the string `[ci skip]`. The default message is defined in controllers/imageupdateautomation_controller.go, as a literal string. You'll want to make it a multiline string, so the `[ci skip]` bit can go in the message part, rather than the summary -- e.g.,

```go
const defaultCommitTemplate = `Update from image update automation

[ci skip]
`
```

Lastly, in _your own_ commit to this PR, please keep the commit summary to 50 characters or fewer; further details can go in the body. See https://github.com/fluxcd/image-automation-controller/blob/main/CONTRIBUTING.md#format-of-the-commit-message and the article linked from there.

fbrv:
Hi @squaremo, thanks for the explanation, reading the issue I was thinking it was related to this repository and not the controller, I'll update the code this evening after I passed the tests

squaremo:
I'm going to close this, as a matter of house-keeping. @fbrv if you do get a chance to have another go at this, just open another PR. Thanks!

PR #73, opened 2021-01-04
Screen files, and output only those updated

### LINKED ISSUE #68: Update of all yaml instead of just the marked line
Hi 👋 

Thank you for your awesome job. I've tested it yesterday and it works like a charm. I've noticed a strange behavior when running automation: every yaml of my repository was rewritten with another formatting.

For exemple, the first commit of `image-automation-controller` on [my repository is this one](https://github.com/vibioh/vith/commit/6914190). The version is [correctly updated](https://github.com/vibioh/vith/commit/6914190#diff-71688e27eea8fddc595d3b07e3e5e72eb4ee1d2c9c0c4cbdb717b07c460f49ffR20) but why the rest also?

I have specific rules for yaml formatting (mainly relying on `prettier`), is there a way (a parameter, a tag) to limit updates to only the lines tagged or only the files which contain tags?

squaremo:
This is intended to address two problems:

 - LocalPackage{Reader,Writer} like to reformat the YAML that passes
   through them; mostly this is harmless, but occasionally it will end
   up fighting format tooling, e.g., `prettier`.

 - It's possible that things like Helm chart templates are lying
   around in the git repository to which automation is applied. Those
   templates have extensions of ".yaml" but are not usually parseable
   as YAML, so would result in errors from the file reader.

This commit changes how updates are run -- firstly, it screens files
by checking for a token (`"$imagepolicy"`) that will be present in
files that might need updating. This cheaply removes some nodes --
likely including Helm chart templates -- from consideration.

Secondly, it now only writes files that were actually updated by an
imagepolicy setter, rather than writing everything that was an
input. This means it's less likely to reformat something that doesn't
need to be touched at all.

Fixes #68.

NB the newly introduced `ScreeningLocalReader` sets aside the name of any files which both contained the token (`"$imagepolicy"`) and were not processable, rather than failing. I'm not making use of that at present, but it could go into an event, or at least the log.


stefanprodan (commented):
Please add the copyright header all go files.

stefanprodan (approved):
LGTM

Thanks @squaremo 🥇

chanwit:
:tada: :tada: :tada:

Issue #70, opened 2020-12-15
GitRepository branch is ignored


stefanprodan:
If no branch is specified in the image automation object, instead of using the branch set in the referenced GitRepository, the controller uses `master` and this breaks the push. We should make the branch required or use the one specified in the GitRepository.

squaremo:
A GitRepository doesn't always have a branch. Rather than come up with a scheme to handle other kinds of ref, I would rather make the branch mandatory in ImageUpdateAutomation.

stefanprodan:
Fixed in #74

PR #74, opened 2021-01-05
Make the branch field mandatory

squaremo:
Instead of having an arbitrary default branch, make the checkout
branch mandatory. This needed a little finessing in the tests, since
they did not cover using different branches (though did cover using a
non-standard branch).


stefanprodan (approved):
LGTM


This fixes: #70

PR #76, opened 2021-01-06
Release v0.2.0

stefanprodan:
This prerelease comes with a fix to the manifest update
mechanism. The controller now only writes files that were
actually updated by an image policy setter instead of 
reformatting all Kubernetes YAMLs.

Starting with this version, the `spec.checkout.branch`
field is mandatory.


squaremo (approved):
Looks good, thanks Stefan 🍊

Issue #71, opened 2020-12-16
Add implementation of libgit2 (Follow configuration from source-controller)


Nordes:
Now on source-controller we can create a `gitrepository` resource for azure using implementation `libgit2`.  Related bug (closed) is https://github.com/fluxcd/source-controller/issues/104

Example:
```bash
flux create source git flux-system \
  --git-implementation=libgit2 \
  --secret-ref=git-ssh-key \
  --url=ssh://git@ssh.dev.azure.com/v3/org/collection/my_flux_repository \
  --branch=main \
  --interval=1m
```

When the image-automation-controller try to get from git, it fails since it's not following the _git-implementation_ directive. 

It would be very nice to be able to have the full flow working. (Adding flux -to- auto-update)

Right now the error is as below:
> {"level":"error","ts":"2020-12-16T18:12:00.877Z","logger":"controller","msg":"Reconciler error","reconcilerGroup":"image.toolkit.fluxcd.io","reconcilerKind":"ImageUpdateAutomation","controller":"imageupdateautomation","name":"flux-system","namespace":"flux-system","error":"unable to clone 'ssh://git@ssh.dev.azure.com/v3/org/collection/my_flux_repository', error: empty git-upload-pack given"}

**Images:**
source-controller:v0.5.4
image-automation-controller:v0.1.0
image-reflector-controller:v0.1.0

phillebaba:
@Nordes this will probably be done as image automation controller reaches an official release. It was a lot of work getting libgit2 in place for source-controller as it still has to work with the original library. Adding it now might probably make the development slower as it reaches the finish line.

I know that you are on vacation @squaremo so whenever you get back and see this can you comment if you have any thoughts on the matter :)

Nordes:
Thanks a lot for the update :). Once it's done, I will be able to upgrade from V1 to V2 ;).

squaremo:
@phillebaba Is there a possibility that only clones need to go through the libgit2 path, while commits and pushes can still be done by go-git?

Nordes:
As the the error is the same I had when the clone was failing (before it was fixed to use go-git). I guess the same protocol is used for push. Which is a bit a pity (Azure Devops).

phillebaba:
So I have spent some time debugging different git clients with Azure Devops and my experience has been that Azure Devops will reject any gitwire protocol request (https and ssh) that does not advertise the correct capabilities in the first handshaking part. So basically I do not think it is possible to commit with go-git.

[...]

PR #82, opened 2021-01-12
Libgit2 support

squaremo:
This adds support for switching to libgit2 for clone and push, for the same reason [source-controller needs that option](https://github.com/fluxcd/source-controller/pull/213).

Fix: #71


.github/workflows/build.yaml
```diff
@@ -25,6 +25,8 @@ jobs:
         go-version: 1.15.x
     - name: Set up kubebuilder
       uses: fluxcd/pkg/actions/kubebuilder@main
+    - name: Install libgit2
+      run: sudo apt-get install -y libgit2-dev
```
stefanprodan, line 29: This will not work as libgit2 is a couple of years old on apt. You can no longer build the binary in the VM, only in container, see https://github.com/fluxcd/source-controller/blob/main/.github/workflows/e2e.yaml#L31

README.md
```diff
@@ -13,3 +13,17 @@ updating YAML files in a git repository, and committing the changes.
 
 Please see the [installation and use
 guide](https://toolkit.fluxcd.io/guides/image-update/).
+
+## How to work on it
+
+You will need to install `libgit2` corresponding to the version of its
```
hiddeco, line 19: Wrote detailed instructions for this for [source-controller](https://github.com/fluxcd/source-controller/blob/main/CONTRIBUTING.md#installing-required-dependencies) from which you may want to piggyback.

hiddeco (approved):
LGTM, but I touched the PR myself so maybe @stefanprodan should have a look too.

Thank you @squaremo :grapes:

stefanprodan (approved):
LGTM

Thanks @squaremo 🍻

PR #84, opened 2021-01-12
Upgrade controller-runtime to v0.7.0

hiddeco:
This commit upgrades the `controller-runtime` dependency to `v0.7.0`,
including all changes required to make all wiring work again.

- Upgrade `runtime` to v0.6.2 to include `controller-runtime` changes.
- Logger has been removed from the reconciler, and is now retrieved
  from the `context.Context` passed to the `Reconcile` method and
  downwards functions.
- Logger configuration flags are now bound to the flag set using
  `BindFlags` from `runtime/logger`, ensuring the same contract across
  GitOps Toolkit controllers, and the `--log-json` flag has been
  deprecated in favour of the `--log-encoding=json` default.
- The `ChangePredicate` from `runtime` has changed to a
  `ReconcileRequestedPredicate`, and is now chained with the
  `GenerationChangedPredicate` from `controller-runtime` using
  `predicate.Or`.
- Signatures that made use of `runtime.Object` have changed to
  `client.Object`, removing the requirement to e.g. call
  `runtime.Object#Object`.
- The `leader-election-role` was changed, as leader election now works
  via the `coordination/v1` API.

Other notable changes:

- Upgrade of `image-reflector-controller` API package to include
  controller-runtime changes.
- Upgrade of `source-controller` API package to v0.6.1.

Ref: https://github.com/fluxcd/flux2/issues/613
Requires: https://github.com/fluxcd/image-reflector-controller/pull/76


api/go.mod
```diff
@@ -6,5 +6,5 @@ require (
 	github.com/fluxcd/pkg/apis/meta v0.5.0
 	k8s.io/api v0.19.3
 	k8s.io/apimachinery v0.19.4
-	sigs.k8s.io/controller-runtime v0.6.3
```
stefanprodan, line 9: k8s.io/api is at 0.19.3, please set it to 0.19.4

controllers/imageupdateautomation_controller.go
```diff
@@ -271,14 +267,11 @@ func durationSinceLastRun(auto *imagev1.ImageUpdateAutomation, now time.Time) ti
...
+func (r *ImageUpdateAutomationReconciler) automationsForGitRepo(obj client.Object) []reconcile.Request {
 	ctx := context.Background()
 	var autoList imagev1.ImageUpdateAutomationList
-	if err := r.List(ctx, &autoList, client.InNamespace(obj.Meta.GetNamespace()), client.MatchingFields{repoRefKey: obj.Meta.GetName()}); err != nil {
-		r.Log.Error(err, "failed to list ImageUpdateAutomations for GitRepository", "name", types.NamespacedName{
-			Name:      obj.Meta.GetName(),
-			Namespace: obj.Meta.GetNamespace(),
-		})
+	if err := r.List(ctx, &autoList, client.InNamespace(obj.GetNamespace()),
+		client.MatchingFields{repoRefKey: obj.GetName()}); err != nil {
```
squaremo, line 274: The log message is removed here. I think it's ultimately OK (other GOTK controllers silently ignore errors in similar situations), I just want to flag it.

squaremo (approved):
Woo thanks for this. It was easy to review because all the changes are carefully considered (and enumerated in the description).

PR #87, opened 2021-01-13
Update to kyaml 0.10.5

squaremo:
This contains a number of fixes, in particular for an indentation bug
that seems to have been introduced not long ago, and which would
otherwise cause problems after updating controller-runtime (which
brings a slightly less recent kyaml with it).

This also comes with a nice enhancement: it's no longer necessary to
use the global schema for kio setters, you can just build a schema and
use that. No need to serialise access to the global schema. Yay!

Signed-off-by: Michael Bridgen <michael@weave.works>

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -25,5 +25,5 @@ require (
 	k8s.io/apimachinery v0.19.4
 	k8s.io/client-go v0.19.4
 	sigs.k8s.io/controller-runtime v0.6.4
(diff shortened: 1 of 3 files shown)


hiddeco (approved):

stefanprodan:
> No need to serialise access to the global schema.

🎉 🎉 🎉

PR #94, opened 2021-01-18
Add API reference for ImageUpdateAutomation

### LINKED ISSUE #54: API user reference for image.toolkit.fluxcd.io/v1alpha1

squaremo:
Fixes #54.

### DIFF
--- a/docs/spec/v1alpha1/imageupdateautomations.md
+++ b/docs/spec/v1alpha1/imageupdateautomations.md
@@ -1,0 +1,195 @@
+<!-- -*- fill-column: 100 -*- -->
+# Image Update Automations
+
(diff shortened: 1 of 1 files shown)


stefanprodan (approved):
LGTM

PR #95, opened 2021-01-20
Reform update strategy types

### LINKED ISSUE #66: `update` field of API is largely empty, and could be optional
The `update` field of the ImageUpdateAutomation spec can only contain one value, since `paths` was removed, and that's `{ setters: {} }`. It isn't defaulted in the schema or in code, so you need to provide

```yaml
# ...
spec:
  update:
    setters: {}
```

which is ugly.

Using `kubebuilder:default` to default it to `{ setters: {} }` doesn't work, because YAML (I think) interprets that inner value as null, and the schema won't accept null.

Assuming we want to keep room for other strategies, a nicer representation might be to use an enum, and (embedded) fields for any parameters:

```yaml
spec:
  update:
    strategy: Setters
    # any strategy parameters go here, either embedded or in an object
```

which means the default value for update can be `{ strategy: "Setters" }`, which is more pleasing.

squaremo:
It's convenient to be able to leave out the update strategy, since
there is only one possible value at present; and if there were
alternatives, the present choice would still be a reasonable
default. However, with the format as it is, this doesn't work with
OpenAPIv3 schema, so you have to supply a value, even though there are
no parameters:

```yaml
spec:
  update:
    setters: {}
```

A more self-explanatory format which _does_ work with defaulting is to
name the strategy rather than relying on the presence of a field:

```yaml
spec:
  update:
    strategy: Setters
```

The whole `update` field can be elided and left to default. This
doesn't preclude having other strategies later, even those with
parameters, e.g.,

```yaml
spec:
  update:
    strategy: Foo
    fooParam: 5
```

This PR changes the API types and code that uses them, and the CRD
manifest, and adds a test that checks the defaulting actually works.

Fixes #66.


squaremo:
Also needs:

 - [ ] doc updates

stefanprodan:
If `setters` was optional there is no need for a conversion webhook, Kubernetes API will drop that field since it's no longer in the CRD spec.

squaremo:
> If `setters` was optional there is no need for a conversion webhook, Kubernetes API will drop that field since it's no longer in the CRD spec.

That's handy! People will still (effectively) have an `update: {}` field; I will test how that affects things when updating the CRDs and controller.

squaremo:
OK, the controller will barf on existing objects once it's updated. To avoid that, I've put a default on the `.spec.update.strategy` field as well. Existing objects will look like `.spec.update: {}` (setters field now ignored), which won't default according to the rule for `.spec.update`, but _will_ default according to the rule for `.spec.update.strategy`, giving the same result.

stefanprodan (approved):
LGTM

PR #115, opened 2021-02-25
Better error messages from `git push`

### LINKED ISSUE #102: Image Automation not providing error message when git-repo has no write-access
We use an ssh-deploy key to have flux talk to repos (we use gitlab.com)

When we don't enable write-access on the git-repo for the deploy-key, flux (correctly) fails to write the image updates, but does not provide a nice error message as to why.

We get the following: (this is also output when using `flux get image update`)

```
message: 'unknown error: remote: '
```

```
apiVersion: v1
items:
- apiVersion: image.toolkit.fluxcd.io/v1alpha1
  kind: ImageUpdateAutomation
  metadata:
    annotations:
    generation: 2
    labels:
      kustomize.toolkit.fluxcd.io/checksum: 0a827fca90af236d1cb0d0e2c757681441c6ccfa
      kustomize.toolkit.fluxcd.io/name: flux-system
      kustomize.toolkit.fluxcd.io/namespace: flux-system
    name: <redacted>
    namespace: flux-system
  spec:
    checkout:
      branch: test-flux2-image-updates
      gitRepositoryRef:
        name:<redacted>
    commit:
      authorEmail: fluxcdbot@users.noreply.github.com
      authorName: fluxcdbot
      messageTemplate: '[ci skip] update image'
    interval: 1m0s
    update:
      setters: {}
  status:
    conditions:
    - lastTransitionTime: "2021-01-28T14:01:47Z"
      message: 'unknown error: remote: '
      reason: ReconciliationFailed
      status: "False"
      type: Ready
    observedGeneration: 2
```

squaremo:
libgit2 and go-git both have flaws in the way they treat errors from the remote. go-git takes only the first line, meaning that it gets a blank error message from GitLab which like to respond with a banner. libgit2 returns the whole response, including blank lines and fences ("=========...").

This commit corrects for both these flaws, by supplying a message if go-git has taken a blank line, and stripping out blank lines and fences from libgit2's error. This is unavoidably a brittle approach, so I have limited it to just the situation that was reported as a problem: pushing to the upstream git repo.

Fixes #102.

### DIFF
--- a/controllers/git_error_test.go
+++ b/controllers/git_error_test.go
@@ -1,0 +1,102 @@
+/*
+Copyright 2020, 2021 The Flux authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package controllers
+
+import (
+	"errors"
+	"testing"
+)
+
+func TestLibgit2ErrorTidy(t *testing.T) {
+	// this is what GitLab sends if the deploy key doesn't have write access
+	gitlabMessage := `remote: 
+remote: ========================================================================
+remote: 
+remote: This deploy key does not have write access to this project.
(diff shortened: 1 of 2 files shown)


controllers/git_error_test.go
```diff
@@ -0,0 +1,86 @@
+package controllers
```
hiddeco, line 1: Test file, but missing the LICENSE header we have everywhere.

hiddeco (approved):
Looks OK to me, tiny nit about a missing license header. :pineapple:

hiddeco (approved):

PR #119, opened 2021-03-02
Supply update result value to the commit message template

squaremo:
The effect of this PR is to give the commit message template some data to work with. Specifically: a data structure (definition given below) which holds the images updated for each object, for each file; and the name of the image automation object that created the commit.

This is the update result:

```go
type Result struct {
	Files map[string]FileResult
}

type FileResult struct {
	Objects map[yaml.ResourceIdentifier][]name.Reference
}
```

.. where the map keys are file paths and `{APIVersion, Kind, Name, Namespace}`, and `name.Reference` is an image name.

Here's a template that uses the result structure to print out what was updated, in different ways:

```go
template := `Commit summary

Image update automation object: {{ .AutomationObject }}

Files:
{{ range $filename, $_ := .Updated.Files -}}
- {{ $filename }}
{{ end -}}

Objects:
{{ range $resource, $_ := .Updated.Objects -}}
- {{ $resource.Kind }} {{ $resource.Name }}
{{ end -}}

Images:
{{ range .Updated.Images -}}
- {{.}}
{{ end -}}
`
```

(There are methods on the Result type to make enumerating the objects and image names possible without descending into the structure.)

EDIT While writing the guide I made the template data more convenient to use (see the commit):
 - [x] Write a reference for the template data
 - [x] Give the name of the policy for each update (A: yes it's useful)

EDIT Not sure whether these are worth it:
 - [?] Wrap ResourceIdentifier in something that has convenient string-returning methods

### DIFF
--- a/controllers/imageupdateautomation_controller.go
+++ b/controllers/imageupdateautomation_controller.go
@@ -69,7 +69,9 @@ const defaultMessageTemplate = `Update from image update automation`
 const repoRefKey = ".spec.gitRepository"
 const imagePolicyKey = ".spec.update.imagePolicy"
 
-type TemplateValues struct {
+// TemplateData is the type of the value given to the commit message
+// template.
+type TemplateData struct {
 	AutomationObject types.NamespacedName
 	Updated          update.Result
 }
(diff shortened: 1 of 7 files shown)


stefanprodan:
After reading the API docs I don't get how to use this feature, is this intended to place a go template inside `messageTemplate`? I think the api doc needs a YAML example.

stefanprodan (approved):
LGTM

Thanks @squaremo 🥇

Issue #120, opened 2021-03-04
Push commits to a specified branch / PR


squaremo:
Discussion about the design:
https://github.com/fluxcd/flux2/discussions/952

squaremo:
Fixed by #121.

PR #121, opened 2021-03-05
Push to branch

squaremo:
This implements the simplest workable design for pushing to a branch other than the branch checked out. If `.spec.push.branch` is present in the `ImageUpdateAutomation` object, the controller will check out that branch, creating it if necessary, and commit to it before pushing to the origin repo.

This is conceptually and mechanically simple, and does enough to build further automation on -- e.g., a GitHub Action that creates a pull request when it sees the branch created.

Addresses #120.


stefanprodan (approved):
LGTM

Thanks @squaremo will you add this to the API spec docs in a different PR?

squaremo:
@stefanprodan Do you think we should mark this as experimental in some way? I feel like I need to try building some e.g., GitHub Actions automation on top to make sure it works nicely in the way we anticipate (but don't have time right now)

stefanprodan:
I'll work on the GH Actions example, if you can update the spec doc to describe the feature it would be great. The whole API is experimental being alpha so I don't think we need to do anything special for this field.

Issue #100, opened 2021-01-24
Restrict update automation to certain folders in a repository


PhilippMT:
Hi there,

I can see there was an "path" field in the spec which was discussed in #24 and removed with pull request #64 due to missing use-cases. I might have a possible use-case. The flux2 documentation states how to setup up multiple clusters/stages in different folders of the same repository. Like the folder "dev" is deployed into the dev-cluster, the folder "prod" to the prod-cluster an so on. But this can  cause conflicts. If different clusters have ImagePolicies with the same name, but different spec, the automation-controller of the dev-cluster will not only adjust the version-tags in the "dev" folder, but also in the folders of the other stages. Possible solutions would be:
1. to use different branches for each stage, or even dedicated repositories but this would limit the use of a "base-kustomization" folder which is reused in all stages
2. to use different policy names in all stages. But there is still the eventuality that a developer accidentally uses a policy name which is already used in an other stage and breaks something.

To implement the "path" field would be a big improvement in this case.

Best regards
Philipp

jvelasquez:
Duplicate of https://github.com/fluxcd/image-automation-controller/issues/72

squaremo:
Not a duplicate of #72, just the similar subject matter. #72 is specifically about whether image-automation-controller should honour the `ignore` field in a `GitRepository` (it's a reasonable expectation; but it raises some tricky questions).

This issue approaches filtering files in the git repository from another direction: what do you do if you want to run two or more different specs for automation on the same git repository? #72 doesn't solve this satisfactorily -- you would have to have `GitRepository` objects pointing at the same git repository with each ignoring all the bits they don't want to automate (which might change, if you e.g., add another cluster). It's a rather indirect way of serving this use case.

stevejr:
I agree that this is going to be a common use case if Flux v2 users are encouraged to build fleet/multi tenant repos. Right now the Image automation is not usable when you have more than 1 cluster with differing image policy specs and the app deployment manifest (be that a HelmRelease, Deployment, Pod etc) stored in a single repo and then using Kustomize to patch accordingly.

[...]

PR #126, opened 2021-03-16
Allow specifying the path for manifests updates

stefanprodan:
Changes:
- Add optional `path` field to `spec.update`, defaults to the git repo root
- Restrict updates to the specified `spec.update.path`

Fix: #100 https://github.com/fluxcd/flux2/discussions/1099

### DIFF
--- a/controllers/update_test.go
+++ b/controllers/update_test.go
@@ -24,6 +24,7 @@ import (
 	"math/rand"
 	"net/url"
 	"os"
+	"path"
 	"path/filepath"
 	"strings"
 	"time"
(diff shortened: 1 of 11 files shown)


stefanprodan:
TODOs after this gets merged and release:
- add `--path` to `flux create image update`
- explain how path can be used to restrict updates for each env in https://toolkit.fluxcd.io/guides/image-update/

squaremo (commented):
Test please! All the bits are there, you could put it alongside the commit message test case.

stefanprodan:
@squaremo I can't run the tests locally, any idea what's going on?

```
Running Suite: Controller Suite
===============================
Random Seed: 1615888184
Will run 19 of 19 specs

2021/03/16 11:49:49 request: GET 127.0.0.1:57077/config-yoh54.git/info/refs?service=git-receive-pack
panic: 
Your test failed.
Ginkgo panics to prevent subsequent assertions from running.
Normally Ginkgo rescues this panic so you shouldn't see it.

But, if you make an assertion in a goroutine, Ginkgo can't capture the panic.
To circumvent this, you should call

        defer GinkgoRecover()

at the top of the goroutine that caused this panic.


goroutine 983 [running]:
github.com/onsi/ginkgo.Fail(0xc000c8c000, 0x13e, 0xc000a079d0, 0x1, 0x1)
        /Users/stefanprodan/go/pkg/mod/github.com/onsi/ginkgo@v1.14.1/ginkgo_dsl.go:260 +0xc8
github.com/onsi/gomega/internal/assertion.(*Assertion).match(0xc000e45f90, 0x5aae958, 0x67ba5d0, 0x0, 0x0, 0x0, 0x0, 0x5912970)
        /Users/stefanprodan/go/pkg/mod/github.com/onsi/gomega@v1.10.2/internal/assertion/assertion.go:79 +0x216
github.com/onsi/gomega/internal/assertion.(*Assertion).ToNot(0xc000563f90, 0x5aae958, 0x67ba5d0, 0x0, 0x0, 0x0, 0xc0006a06c0)
        /Users/stefanprodan/go/pkg/mod/github.com/onsi/gomega@v1.10.2/internal/assertion/assertion.go:43 +0xc7
github.com/fluxcd/image-automation-controller/controllers.glob..func1.1(0xc000a01ad0)
        /Users/stefanprodan/go/src/github.com/fluxcd/image-automation-controller/controllers/suite_test.go:94 +0xf9
created by github.com/fluxcd/image-automation-controller/controllers.glob..func1
        /Users/stefanprodan/go/src/github.com/fluxcd/image-automation-controller/controllers/suite_test.go:92 +0x9d6
FAIL    github.com/fluxcd/image-automation-controller/controllers       4.934s
ok      github.com/fluxcd/image-automation-controller/pkg/test  0.263s  coverage: 71.2% of statements
ok      github.com/fluxcd/image-automation-controller/pkg/update        0.189s  coverage: 85.9% of statements
FAIL
make: *** [test] Error 1

```

controllers/update_test.go
```diff
@@ -263,6 +264,113 @@ Images:
...
+						Path:     "./overlay",
+					},
+					Commit: imagev1.CommitSpec{
+						MessageTemplate: "nothing",
+					},
+				},
+			}
+			Expect(k8sClient.Create(context.Background(), updateBySetters)).To(Succeed())
+			// wait for a new commit to be made by the controller
+			waitForNewHead(localRepo, branch)
```
squaremo, line 366: The main thing added in this PR is to restrict updates to a path; as a demonstration that it works I would expect to see a file that _could_ be updated without the new restriction, but isn't. You could e.g., put a file with the same marker _outside_ the path, and check afterwards that it's not updated (and that the one in the path is). `compareRepoWithExpected` helps with the last bit -- there's examples of using it lower in the file.

stefanprodan: Found a better way to check the restriction, I couldn't use `compareRepoWithExpected` as it is.

squaremo: Works for me :+1:

squaremo (approved):
All looks good, thanks Stefan 🥭

nomeelnoj:
I know it's a bit late, but would it be possible to add globbing to the path object? This would help handle the use case where resources are put into folders as follows:

```
apps/myapp/base/kustomization.yaml
apps/myapp/overlays/dev/deployment.yaml # image policy marker found here
apps/myapp/overlays/stg/deployment.yaml # image policy marker found here
```
Then, setting:
```
path: ./apps/**/dev
```
in the `ImageUpdateAutomation` would allow this folder structure to be supported, while allowing a single `ImagePolicy` object to be shared across all the workloads.

therapy-lf:
> I know it's a bit late, but would it be possible to add globbing to the path object? This would help handle the use case where resources are put into folders as follows:
> 
> ```
> apps/myapp/base/kustomization.yaml
> apps/myapp/overlays/dev/deployment.yaml # image policy marker found here
> apps/myapp/overlays/stg/deployment.yaml # image policy marker found here
> ```
> 
> Then, setting:
> 
> ```
> path: ./apps/**/dev
> ```
> 
> in the `ImageUpdateAutomation` would allow this folder structure to be supported, while allowing an `ImagePolicy` object with the same name (different objects in each cluster, with different policies, but each named `myapp`, e.g.) to be shared across all the workloads.

This will be very useful. Would be great if you implement this feature.

PR #130, opened 2021-03-18
Fix typo in template code example

jasonraimondi:
Fixes a minor typo that causes CRD violation if copy and pasted

```
error: error validating "./clusters/staging/flux-system/image-update-automation.yml": error validating data: ValidationError(ImageUpdateAutomation.spec.commit): unknown field "messsageTemplate" in io.fluxcd.toolkit.image.v1alpha1.ImageUpdateAutomation.spec.commit; if you choose to ignore these errors, turn validation off with --validate=false
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,7 +17,7 @@ images that were updated along with the resource kind and name e.g.:
 ```yaml
 spec:
   commit:
-    messsageTemplate: |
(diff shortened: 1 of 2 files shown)


stefanprodan (approved):
LGTM

Thanks @jasonraimondi good catch!

Issue #131, opened 2021-03-19
Controller logs resource conflict error messages


squaremo:
{"level":"error","ts":"2021-03-19T15:34:29.949Z","logger":"controller-runtime.manager.controller.imageupdateautomation","msg":"Reconciler error","reconciler group":"image.toolkit.fluxcd.io","reconciler kind":"ImageUpdateAutomation","name":"flux-system","namespace":"flux-system","error":"Operation cannot be fulfilled on imageupdateautomations.image.toolkit.fluxcd.io \"flux-system\": the object has been modified; please apply your changes to the latest version and try again"}

The line at which this problem arises is https://github.com/fluxcd/image-automation-controller/blob/main/controllers/imageupdateautomation_controller.go#L263.

bigkevmcd:
I've got a fix lined up for this, if I can work out a test

stefanprodan:
More users are hitting this: https://cloud-native.slack.com/archives/CLAJ40HV3/p1616283160302500?thread_ts=1616279639.295100&cid=CLAJ40HV3

PR #132, opened 2021-03-22
Update ImageUpdateAutomation Status with Patch.

bigkevmcd:
This changes the functionality when updating the status to use Patch
rather than Update, which is more resilient to changes.

Fix: #131

### DIFF
--- a/controllers/imageupdateautomation_controller.go
+++ b/controllers/imageupdateautomation_controller.go
@@ -127,7 +127,8 @@ func (r *ImageUpdateAutomationReconciler) Reconcile(ctx context.Context, req ctr
 	// annotation if it's there
 	if token, ok := meta.ReconcileAnnotationValue(auto.GetAnnotations()); ok {
 		auto.Status.SetLastHandledReconcileRequest(token)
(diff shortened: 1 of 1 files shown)


stefanprodan (approved):
LGTM

Thanks @bigkevmcd ☕

controllers/imageupdateautomation_controller.go
```diff
@@ -291,6 +291,21 @@ func (r *ImageUpdateAutomationReconciler) SetupWithManager(mgr ctrl.Manager) err
 		Complete(r)
 }
 
+func (r *ImageUpdateAutomationReconciler) patchStatus(ctx context.Context,
+	req ctrl.Request,
+	newStatus imagev1.ImageUpdateAutomationStatus) error {
+
+	var auto imagev1.ImageUpdateAutomation
+	if err := r.Get(ctx, req.NamespacedName, &auto); err != nil {
```
squaremo, line 299: This can set the status of a different version of the object than the one we're reconciling. For that reason, I think this is a fragile fix, even if it doesn't cause other problems.

squaremo:
I would like to have at least seen a test before this was merged.

PR #136, opened 2021-03-24
Enable GPG Signing of Commits

### LINKED ISSUE #122: [feature] Commit with gpg signing
Currently documentation shows how to utilize signing for secrets but that same gpg key can be used for commits as well.
Need to add `signingkey` to the spec.

LWJ:
Closes #122 

Added a new (optional) SigningKey field to the CommitSpec with a secretRef. This secret can be generated the exact same way as described in the [SOPS guide](https://toolkit.fluxcd.io/guides/mozilla-sops/#generate-a-gpg-key), so it should be easy for existing users to understand - the only difference is the key is named `git.asc` rather than  `sops.asc`.

Once any modifications have been suggested/made, I could make a pull request to the flux2 project with some documentation and CLI changes.


squaremo (changes requested):
Requested changes:

 - make a const for `"git.asc"`
 - change the field `secretRef` to a value type, since it's required (and could cause an NPE otherwise)

api/v1alpha1/imageupdateautomation_types.go
```diff
@@ -142,6 +145,16 @@ type ImageUpdateAutomationStatus struct {
...
 }
 
+// SigningKey references a Kubernetes secret that contains a GPG keypair
+type SigningKey struct {
+	// SecretRef holds the name to a secret that contains a 'git.asc' key
+	// corresponding to the ASCII Armored file containing the GPG signing
+	// keypair as the value. It must be in the same namespace as the
+	// ImageUpdateAutomation.
+	// +required
+	SecretRef *meta.LocalObjectReference `json:"secretRef,omitempty"`
```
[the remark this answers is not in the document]
LWJ: It definitely could be reduced. The reason I added the nesting is to try and be consistent with the other flux CRDs which appear to always use `secretRef`, and enclosing in a wrapping type where appropriate. For example, the `KubeConfig` in the HelmRelease CRD: https://github.com/fluxcd/helm-controller/blob/main/api/v2beta1/helmrelease_types.go#L205

Happy to collapse the field in this case if you think it's appropriate though.

LWJ:
Thanks @squaremo for reviewing! I'll get these changes made this evening :smile:

api/v1alpha1/imageupdateautomation_types.go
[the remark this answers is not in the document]
squaremo: Consistency is a good instinct :+1: Let's leave as is.

controllers/update_test.go
```diff
@@ -388,6 +390,162 @@ Images:
 		})
 	})
 
+	Context("commit signing", func() {
+
+		var localRepo *git.Repository
+
+		// generate keypair for signing
+		pgpEntity, err := openpgp.NewEntity("", "", "", nil)
+		Expect(err).ToNot(HaveOccurred())
```
squaremo, line 399: This won't work as you expect outside of BeforeEach/AfterEach/It. Make pgpEntity a variable and initialise it in `BeforeEach`.

squaremo: I'm not quite sure how the test still passes -- maybe because it doesn't matter how it gets initialised, so long as it does once. In any case, it would be neater to keep initialisation inside `BeforeEach`.

LWJ: Good catch, have shifted this around

squaremo (approved):
This all looks good, nice work @LWJ 🍎

PR #139, opened 2021-03-29
Add v1alpha2 API version

squaremo:
This adds a v1alpha2 API, without changing the controller or tests to
use it (yet). The new API has roughly the desired shape, per
https://github.com/fluxcd/flux2/discussions/1124, but supporting only
things that the controller supports now.

TODO:

 - [x] `.spec.sourceRef` rather than `.spec.checkout.gitRepositoryRef`
 - [x] move git-specific fields to `.spec.git`
 - [x] `.spec.git.commit.author.{email,user}`
 - [x] `.spec.git.checkout.ref` rather than `...branch`
 - [x] generate spec reference doc from v1alpha2
 - [x] rewrite spec guide
 - [x] write a guide for rewriting YAMLs from v1alpha1 to v1alpha2 

EDIT: I put the migration guide at the bottom of the v1alpha2 spec guide.


stefanprodan (approved):
LGTM

Thanks @squaremo 🏅

squaremo:
Should this also remove the v1alpha1 types and CRD definition @stefanprodan?

stefanprodan:
I think we need to test what happens with and without v1alpha1 after an upgrade if the manifests in the repo are on v1alpha1. For the other controllers we removed the alpha version when we switched to beta but this is different as we are moving from one alpha to another.

squaremo:
> I think we need to test what happens with and without v1alpha1 after an upgrade if the manifests in the repo are on v1alpha1.

If you have a `v1alpha1` `ImageUpdateAutomation` object, and you apply a CRD with _both_ `v1alpha1` and `v1alpha2` definitions, then Kubernetes attempts to convert your object to `v1alpha2` and loses most of the fields. I tried this by applying the CRD YAML with `kubectl`, then fetching the object with `kubectl get imageupdateautomation -o yaml`. I presume a controller using `v1alpha2` would see the same thing.

The other situation is if you have a v1alpha1 object and apply the CRD with _just_ the v1alpha2 definition. When I apply a CRD which defines only v1alpha2, it fails with the error:

    The CustomResourceDefinition "imageupdateautomations.image.toolkit.fluxcd.io" is invalid: status.storedVersions[0]: Invalid value: "v1alpha1": must appear in spec.versions

.. which makes me think applying it as-is in a `flux install` / `bootstrap` will make it fail.

PR #142, opened 2021-04-02
Ignore broken symlinks and outside path, in commit

### LINKED ISSUE #135: Empty commits
I have an ImageUpdateAutomation object with spec.interval set to 1m0s. I see a new commit to my repo every minute, whether or not any image has changed, so the vast majority of these commits are empty commits (i.e. no changes).

The empty commits are, I suppose, harmless; but it would be nice if there was a way of avoiding them.

squaremo:
There is a bug in go-git which leads to it reporting broken, absolute
symlinks as modified whether they are or not:

    https://github.com/go-git/go-git/issues/253

To date, the controller checks whether the repo it has run an update
on is Clean, and as a consequence will run into the bug above if a
broken symlink is in the repo. The result is that it makes and pushes
an empty commit every interval.

To work around the problem, this commit adds a more careful check of
the repo status. Each file reported as modified is validated, firstly
by checking that it's under the path given; and secondly by checking
specifically that it's not a broken symlink.

For convenience, I have moved a few procedures around so they can be
used more readily by go tests.

Fixes #135.

----

> firstly by checking that it's under the path given

Of this bit I am not convinced. It's possible you have a valid symlink under the path, targeting a file _not_ under the path, and you want the update to work. It's also possible you've come to this arrangement by accident -- but I shouldn't presume so!

UPDATE: I removed the path limitation in another commit; it can be squashed in, or reset out.

### DIFF
--- a/controllers/git_test.go
+++ b/controllers/git_test.go
@@ -1,0 +1,107 @@
+package controllers
+
+import (
+	"io/ioutil"
+	"os"
+	"path/filepath"
+	"testing"
+	"time"
+
+	"github.com/go-git/go-billy/v5/memfs"
+	"github.com/go-git/go-git/v5"
+	"github.com/go-git/go-git/v5/plumbing/object"
+	"github.com/go-git/go-git/v5/storage/memory"
+)
+
+func populateRepoFromFixture(repo *git.Repository, fixture string) error {
+	working, err := repo.Worktree()
+	if err != nil {
+		return err
+	}
+	fs := working.Filesystem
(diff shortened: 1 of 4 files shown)


stefanprodan (approved):
LGTM

Thanks @squaremo 🥇 

@hiddeco please incorporate the symlink check in https://github.com/fluxcd/flux2/pull/968

squaremo:
I'm going to squash out the commit removing the path limitation, then merge.

Issue #85, opened 2021-01-13
ImageUpdateAutomation in different namespace than ImagePolicy


TrueBrain:
I have the following setup:

Flux2 is running in namespace `flux-system`.
Prometheus is running in namespace `monitoring`.

I am trying to get image automation to work with as little duplication as possible.

So, I defined `ImagePolicy` (and `ImageRepository`) close to Prometheus. This means, because I use kustomize, that it is forced in the `monitoring` namespace.
Next, I defined `ImageUpdateAutomation` in the `flux-system` namespace, as it wants access to the `GitRepository` which is already defined there.

Now my problem: this setup doesn't work when I try `# {"$imagepolicy": "monitoring:prometheus:tag"}`.

From what I understand, and what does work:
- either `ImagePolicy` needs to be in the `flux-system` namespace (which is far from ideal from a Infrastructure-as-Code point of view)
- or I need to move `ImageUpdateAutomation` into `monitoring` namespace. I don't really have an issue with that, except for the minor code duplication, but the `gitRepositoryRef` doesn't accept `namespace`. That would mean I also have to duplicate `GitRepository`, which would be very prune to errors.

Now I am not sure that what I am trying here is how this is meant to be used, or that I am overlooking some setting somewhere. So if you have any pointers for me how I should be doing this, or if there is an option I can use to make this work, that would be very welcome :)

Thank you!

stefanprodan:
I do think we should use make use of `sourceRef` in the same way we do in kustomize-controller and helm-controller APIs e.g. [CrossNamespaceSourceReference](https://github.com/fluxcd/kustomize-controller/blob/main/api/v1beta1/reference_types.go#L41), where you can refer to a GitRepository from a different namespace. 

PS. This is my personal view, I think @squaremo would disagree.

TrueBrain:
> 
> Can you explain why this is less than ideal?
> 
> Are your sync definitions (`Kustomization` etc.) in `flux-system` with the `GitRepository` objects? I think it's reasonable -- preferable, even -- to run automation in the same place as syncs. This leads me to the alternative quoted above: the ImageRepository and ImagePolicy objects are for automation, so get segregated into a system namespace, i.e., `flux-system`.
> 

Yes, I could have used a bit more words to explain my point, sorry about that :)

From a functional perspective, I fully get what you mean. The problem for me is when you look at this from a Infrastructure as Code perspective. To give an example:

In `kustomization.yaml`, I would need to define an image like this:

```
[...]

PR #151, opened 2021-04-15
Load image policies from all namespaces.

vladlosev:
This PR allows the image update automation to apply image update policies from namespaces other than its own, as a quick way of implementing https://github.com/fluxcd/image-automation-controller/issues/85.  This is done by loading the policies for consideration from all namespaces rather than from just the automation's own namespace.  

In our company we have multiple semi-independent teams working on different
projects in our cluster and a single infrastructure team managing the
infrastructure for them, including the cluster itself.  For the purposes of
data protection we have to keep the teams isolated, so each team has a
namespace it controls, and the infrastructure team runs the system namespaces,
including `flux-system`, where the repository and image automation CRDs are
defined.  The individual teams control the exact set of images they generate
and push and the image selection criteria for deployment so we want to allow
them to manage the `ImageRepository` and `ImageUpdatePolicy` in their
namespaces. But with the `GitRepository` and `ImageUpdateAutomation` residing
in `flux-system`, they can't, unless ImageUpdateAutomation can handle policies
from other namespaces.

### DIFF
--- a/controllers/imageupdateautomation_controller.go
+++ b/controllers/imageupdateautomation_controller.go
@@ -213,10 +213,10 @@ func (r *ImageUpdateAutomationReconciler) Reconcile(ctx context.Context, req ctr
 	switch {
 	case auto.Spec.Update != nil && auto.Spec.Update.Strategy == imagev1.UpdateStrategySetters:
 		// For setters we first want to compile a list of _all_ the
-		// policies in the same namespace (maybe in the future this
+		// policies in all namespaces (maybe in the future this
 		// could be filtered by the automation object).
 		var policies imagev1_reflect.ImagePolicyList
-		if err := r.List(ctx, &policies, &client.ListOptions{Namespace: req.NamespacedName.Namespace}); err != nil {
+		if err := r.List(ctx, &policies, &client.ListOptions{}); err != nil {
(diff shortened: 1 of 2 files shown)


stefanprodan:
@vladlosev thanks for taking the time to open this PR but it’s in current form we can’t accept such a change that would break everyone’s setups. A better way of doing this, without changing the default behavior would be to introduce a namespace label selector in the API spec.

vladlosev:
Hmm, I see what you mean - you want explicit control over which `ImageUpdateAutomation` applies any given `ImageUpdatePolicy` and you lose that control with this patch. Too bad I didn't catch that issue earlier. :-(

I could try my hand at the API change. When you talk about the API change are you talking about one of the two approaches mentioned in https://github.com/fluxcd/image-automation-controller/issues/85#issuecomment-802152827? If so, which one?

vladlosev:
Closing in favor of #157.

squaremo:
@vladlosev Would it also work to define the `ImageUpdateAutomation` object in the user's namespace, and be able to refer to a `GitRepository` defined centrally?

vladlosev:
@squaremo certainly. I'd be happy to use cross-namespace `GitRepository` references.

PR #162, opened 2021-05-06
Fix image name marker

relu:
The image name was incorrectly extracted from the given reference,
yielding a trimmed version of the string and breaking the resulting
replacement value.

For example, if the policy latest image was: `ghcr.io/stefanprodan/podinfo:5.2.0`
It would yield: `ghcr.i`.


stefanprodan (approved):
LGTM

Thanks @relu 🏅

PR #169, opened 2021-05-26
Update source-controller to v0.13.0

hiddeco:
To pull in Git package changes from
https://github.com/fluxcd/source-controller/pull/365.

Ref: https://github.com/fluxcd/flux2/issues/1443


stefanprodan (approved):
🚀

Issue #172, opened 2021-05-31
flux 0.14.1 empty git-upload-pack given


rayanebel:
Hi everyone,

Today I've upgraded our flux instance to the latest version `0.14.1` and after migrating all our `ImageUpdateAutomation` resources to the new format according to this [documentation](https://github.com/fluxcd/flux2/discussions/1333), we have the following error message: `empty git-upload-pack given` which is unclear for us.

We are using Github and we have private repositories.

```bash
NAMESPACE  	NAME                                        	READY	MESSAGE                    	LAST RUN	SUSPENDED
flux-system	imageupdateautomation/advisors              	False	empty git-upload-pack given	        	False
flux-system	imageupdateautomation/client360             	False	empty git-upload-pack given	        	False
flux-system	imageupdateautomation/transactions          	False	empty git-upload-pack given	        	False
flux-system	imageupdateautomation/users                 	False	empty git-upload-pack given	        	False
```

You can find below the yaml of one of our `ImageUpdateAutomation` resource

```yaml
apiVersion: image.toolkit.fluxcd.io/v1alpha2
kind: ImageUpdateAutomation
metadata:
  name: products
  namespace: flux-system
spec:
  git:
    checkout:
      ref:
        branch: main
    commit:
      author:
        email: fluxcdbot@users.noreply.github.com
        name: fluxcdbot
      messageTemplate: '{{range .Updated.Images}}{{println .}}{{end}}'
    push:
      branch: main
  interval: 1m0s
  sourceRef:
    kind: GitRepository
    name: flux-system
  update:
    path: ../applications/apis/overlays/dev
    strategy: Setters
```

Can someone help us to understand what's going on and what's this error message means ?

Thanks.

stefanprodan:
Can you post here the GitRepository manifest, the relevant part is the Git URL.

rayanebel:
Hi @stefanprodan 

Here you can find the `flux-system` GitRepository manifest.

```yaml
apiVersion: source.toolkit.fluxcd.io/v1beta1
kind: GitRepository
metadata:
  name: flux-system
  namespace: flux-system
spec:
  gitImplementation: go-git
  interval: 1m0s
  ref:
    branch: main
  secretRef:
    name: flux-system
  timeout: 20s
  url: https://github.com/xxxxxx/fred-gcp-gitops.git
```

stefanprodan:
Can you please test with `gitImplementation: libgit2` and see if it works? thanks

[...]

PR #173, opened 2021-05-31
Use patched go-git version

hiddeco:
To work around `object not found` errors due to a newly introduced bug
in go-git v5.4.0.

Image available for testing: `hiddeco/image-automation-controller:go-git-patch-776e182`

May contain fix for #172

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -4,6 +4,9 @@ go 1.15
 
 replace github.com/fluxcd/image-automation-controller/api => ./api
 
+// Patch from: https://github.com/go-git/go-git/pull/326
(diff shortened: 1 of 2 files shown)


stefanprodan:
To test this on a cluster bootstrapped with Flux, change the image in your `./clusters/<name>/flux-system/kustomization.yaml` like so:

```yaml
apiVersion: kustomize.config.k8s.io/v1beta1
kind: Kustomization
resources:
  - gotk-components.yaml
  - gotk-sync.yaml
images:
- name: ghcr.io/fluxcd/image-automation-controller
  newName: docker.io/hiddeco/image-automation-controller
  newTag: go-git-patch-776e182
```

stefanprodan:
Closing in favour of: #174

Issue #176, opened 2021-06-02
v0.10.0 Updates unsuccessful with both implementations


Robbilie:
I currently have 2 gitrepository resources in the cluster. they are both based on repos in a bitbucket server. they both sync perfectly fine.

about a week ago the flux version was updated and thus came 0.10.0 of the automation controller.

since then, the controller logs the following error:

go-git:

{"level":"error","ts":"2021-06-02T10:41:14.166Z","logger":"controller-runtime.manager.controller.imageupdateautomation","msg":"Reconciler error","reconciler group":"image.toolkit.fluxcd.io","reconciler kind":"ImageUpdateAutomation","name":"demo","namespace":"services","error":"empty git-upload-pack given"}

libgit2:

{"level":"error","ts":"2021-06-02T09:06:04.731Z","logger":"controller-runtime.manager.controller.imageupdateautomation","msg":"Reconciler error","reconciler group":"image.toolkit.fluxcd.io","reconciler kind":"ImageUpdateAutomation","name":"demo","namespace":"services","error":"unable to clone 'https://xxx/bb/scm/sulu/sulu_services_config', error: failed to connect to xxx: Operation timed out"}

any ideas where this comes from?

stefanprodan:
Duplicate of:  #172 

A fix will be delivered today

squaremo:
> Duplicate of: #172

I don't see a timeout when using libgit2 mentioned in #172.

stefanprodan:
@squaremo I guess that timeout is due to libgit2 not doing shallow cloning.

@Robbilie you can increase the timeout with `GitRepository.spec.timeout`, set it to something like 60s, I guess your repo is quite large.

Robbilie:
i just rolled back to 0.9.1 of the controller and apparently it doesnt work with it either, not sure how it ever worked

ill try the timeout :)

Robbilie:
The repo is actually quite small tbh, thats why im really confused. the bitbucket logs also show 200 status codes for all requests which is also weird…

stefanprodan:
@Robbilie there is a new image available that you could try: `ghcr.io/fluxcd/image-automation-controller:v0.10.1` this should fix the go-git bug. Please let me know if it works for you.

Robbilie:
i just tried libgit2 after i fixed a colleague having inserted the wrong update marker -.-

i now tried your new image, but it still throws the same error with go-git, libgit2 works though

stefanprodan:
[...]

PR #177, opened 2021-06-02
Use libgit2 for clone, fetch, push

### LINKED ISSUE #164: Problem with custom
I've got the basic image update automation configuration set up:

```
apiVersion: image.toolkit.fluxcd.io/v1alpha2
kind: ImageUpdateAutomation
metadata:
  name: flux-system
  namespace: flux-system
spec:
  git:
    checkout:
      ref:
        branch: main
    push:
      branch: pending
    commit:
      author:
        email: paul@paullockaby.com
        name: fluxcd
      messageTemplate: '{{range .Updated.Images}}{{println .}}{{end}}'
  interval: 1m0s
  sourceRef:
    kind: GitRepository
    name: flux-system
  update:
    path: ./flux-development
    strategy: Setters
```

It pushes to GitHub on a branch called "pending". If I merge "pending" but do not _delete_ pending then I end up with this error message in my logs, repeating forever. As long as this error message is happening automation updates stop happening The moment that I delete the "pending" branch from GitHub the error message goes away and automation resumes.

```
{"level":"error","ts":"2021-05-11T00:21:28.837Z","logger":"controller-runtime.manager.controller.imageupdateautomation","msg":"Reconciler error","reconciler group":"image.toolkit.fluxcd.io","reconciler kind":"ImageUpdateAutomation","name":"flux-system","namespace":"flux-system","error":"object not found"}
```

I'd also like to ask if the image automation will automatically detect changes to the image even when the tag doesn't change. For example, if I'm deploying haproxy:2.2.14 the maintainers will periodically rebuild their image when their base image changes (e.g. they rebuild 2.2.14 because of a CVE in a dependent library like SSL) and they'll release a new "version" of 2.2.14 with the same exact tag but with a different hash. Can/does flux2 pick that up and deploy the new image?

squaremo:
source-controller/pkg/git does shallow clones when using the go-git
implementation, and apparently this causes problems when fetching a
branch that has been merged at the origin:

    https://github.com/fluxcd/image-automation-controller/issues/164

So far as I can tell, getting a shallow clone breaks the automation,
no matter whether go-git or libgit2 is used for operations after
cloning. So: just use libgit2 for cloning, which means non-shallow
clones; and, for fetch and push, since there's no functional
difference between the implementations for those.

Fixes #164.

 - [ ] alter docs to make it clear that the git implementation is ignored


stefanprodan (approved):
LGTM

This should solve #176 as we no longer use go-git for push operations.

Thanks @squaremo 🥇

hiddeco (approved):

PR #178, opened 2021-06-02
Release v0.10.2

stefanprodan:


hiddeco (commented):
This is technically a behavioral change and would require a MINOR bump, but I can totally see why you do not want to do this in a way.

On the other hand, we have excluded the automation controllers from the defining `flux` SemVer range, which means this can be MINOR bumped without doing the same for the CLI.

Issue #140, opened 2021-03-30
controller crash when patching YAMLs with non-ASCII characters


ywyt738:
image-automation-controller version 0.6.1 and 0.7.0 also have this issue.
image-automation-controller crashed when the ImageUpdateAutomation is created.

```yaml
apiVersion: image.toolkit.fluxcd.io/v1alpha1
kind: ImageUpdateAutomation
metadata:
  name: imageupdate
spec:
  checkout:
    branch: master
    gitRepositoryRef:
      name: default
  commit:
    authorEmail: fluxcdbot@users.noreply.porsche.com
    authorName: fluxcdbot
    messageTemplate: |
      [ci skip]Automated image update

      Automation name: {{ .AutomationObject }}

      Files:
      {{ range $filename, $_ := .Updated.Files -}}
      - {{ $filename }}
      {{ end -}}

      Objects:
      {{ range $resource, $_ := .Updated.Objects -}}
      - {{ $resource.Kind }} {{ $resource.Name }}
      {{ end -}}

      Images:
      {{ range .Updated.Images -}}
      - {{.}}
      {{ end -}}
  interval: 1m0s
  update:
    strategy: Setters
```

The image-automation-controller log
```bash
{"level":"info","ts":"2021-03-30T06:18:16.540Z","logger":"controller-runtime.metrics","msg":"metrics server is starting to listen","addr":":8080"}
{"level":"info","ts":"2021-03-30T06:18:16.541Z","logger":"setup","msg":"starting manager"}
I0330 06:18:16.542567       7 leaderelection.go:243] attempting to acquire leader lease flux-system/79628f79.fluxcd.io...
{"level":"info","ts":"2021-03-30T06:18:16.543Z","logger":"controller-runtime.manager","msg":"starting metrics server","path":"/metrics"}
I0330 06:18:32.118539       7 leaderelection.go:253] successfully acquired lease flux-system/79628f79.fluxcd.io
{"level":"info","ts":"2021-03-30T06:18:32.118Z","logger":"controller-runtime.manager.controller.imageupdateautomation","msg":"Starting EventSource","reconciler group":"image.toolkit.fluxcd.io","reconciler kind":"ImageUpdateAutomation","source":"kind source: /, Kind="}
[...]

PR #182, opened 2021-06-09
Update go-yaml with changes to indentation style

### LINKED ISSUE #92: kyaml upstream bugs
This is an issue to raise awareness that kustomize kyaml has some serious bugs that could affect Flux users:

- If you have non-ASCII characters in your YAMLs the controller will panic and crash loop https://github.com/kubernetes-sigs/kustomize/issues/3605 
- kyaml: fatal error: concurrent map read and map write https://github.com/kubernetes-sigs/kustomize/issues/3659
- If you have non-string keys in your manifests, the image-automation-controller will not be able to patch the image tags, ref: https://github.com/kubernetes-sigs/kustomize/issues/3446
- YAML anchors and aliases return error on List types https://github.com/kubernetes-sigs/kustomize/issues/3614
- If you have duplicate keys in your manifests, the image-automation-controller will crash loop, ref: https://github.com/kubernetes-sigs/kustomize/issues/3659 (**fixed**)

Since  image-automation-controller depends on kyaml setters, we can't just disable it like we did for [kustomize-controller](https://github.com/fluxcd/kustomize-controller/pull/241).

stefanprodan:
This PR updates the YAML packages (used by the controller to patch Kubernetes manifests) to fix long standing issues like panics that make the controller unusable for many users.

Dependencies changes:
- update `gopkg.in/yaml.v3` from  `v3.0.0-20200313102051-9f266ea9e77c` to `v3.0.0-20200615113413-eeeca48fe776` (fix: #140)
- update `sigs.k8s.io/kustomize/kyaml` from `v0.10.19` to `v0.10.21`

### Breaking change

The `gopkg.in/yaml.v3` update means that the indentation style changed:

From:

```yaml
spec:
  containers:
  - name: one
    image: image1:v1.0.0 # {"$imagepolicy": "automation-ns:policy1"}
  - name: two
    image: image2:v1.0.0 # {"$imagepolicy": "automation-ns:policy2"}
```

To:

```yaml
spec:
  containers:
    - name: one
      image: image1:v1.0.0 # {"$imagepolicy": "automation-ns:policy1"}
    - name: two
      image: image2:v1.0.0 # {"$imagepolicy": "automation-ns:policy2"}
```

While Kustomize kyaml is using an older version of go-yaml we can't hold onto this forever. I find it unacceptable for Flux to keep crashing due to bugs fixed upstream in go-yaml. Kustomize may chose to delay this update (see https://github.com/kubernetes-sigs/kustomize/pull/3789 and https://github.com/kubernetes-sigs/kustomize/issues/3946) but this doesn't mean we are obliged to do the same. The image automation is alpha and breaking changes like this are acceptable considering the tradeoffs. I hope at some point, go-yaml will be able to detect and preserve the original indentation but it seems unlikely https://github.com/go-yaml/yaml/issues/661#issuecomment-733677290

Closes #92

### DIFF
--- a/controllers/testdata/appconfig-expected/deploy.yaml
+++ b/controllers/testdata/appconfig-expected/deploy.yaml
@@ -6,5 +6,5 @@ spec:
   template:
     spec:
       containers:
-      - name: hello
-        image: helloworld:1.0.1 # SETTER_SITE
+        - name: hello
+          image: helloworld:1.0.1 # SETTER_SITE
--- a/controllers/testdata/appconfig-expected2/deploy.yaml
+++ b/controllers/testdata/appconfig-expected2/deploy.yaml
@@ -6,5 +6,5 @@ spec:
   template:
     spec:
       containers:
-      - name: hello
-        image: helloworld:1.2.0 # SETTER_SITE
+        - name: hello
+          image: helloworld:1.2.0 # SETTER_SITE
--- a/controllers/testdata/appconfig-setters-expected/deploy.yaml
+++ b/controllers/testdata/appconfig-setters-expected/deploy.yaml
@@ -6,5 +6,5 @@ spec:
   template:
     spec:
       containers:
-      - name: hello
-        image: helloworld:1.0.1 # SETTER_SITE
+        - name: hello
+          image: helloworld:1.0.1 # SETTER_SITE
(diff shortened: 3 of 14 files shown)


hiddeco (approved):
> While Kustomize kyaml is using an older version of go-yaml we can't hold onto this forever. I find it unacceptable for Flux to keep crashing due to bugs fixed upstream in go-yaml.

I do agree, and think that the number of issues it resolves outweigh the breaking change introduces.

:+1: from me.

squaremo (approved):
Meta-approval (I trust the other reviewer(s)): I am fine with the breaking change, so long as we flag it prominently, of course.

PR #183, opened 2021-06-09
Update dependencies

hiddeco:
- Go 1.16
- `sigs.k8s.io/controller-runtime` to `v0.9.0`
- `fluxcd/source-controller` to `v0.14.0`
- `fluxcd/pkg/*` to packages compatible with Go 1.16 and
   controller-runtime v0.9.0
- image-reflector-controller to `v0.10.0`

Ref: https://github.com/fluxcd/flux2/issues/1486


api/v1alpha2/groupversion_info.go
```diff
@@ -14,9 +14,6 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-// Package v1alpha2 contains API Schema definitions for the image v1alpha2 API group
```
squaremo, line 17: Are these unnecessary?

hiddeco: No, they are already declared in `v1alpha2/doc.go`.

stefanprodan (approved):

api/v1alpha2/groupversion_info.go
squaremo: Got it. I seem to always end up in a situation where kubebuilder's markers are missing or in the wrong place, so I am over-vigilant about markers :-S

Issue #186, opened 2021-06-15
"Unable to extract public key from private key" for ed25519


LalitMaganti:
If an ed25519 SSH key is created during `flux bootstrap` adding a `ImagePolicy` causes the following error to be producer by the image-automation-contoller:
```
{"level":"error","ts":"2021-06-15T22:29:20.579Z","logger":"controller-runtime.manager.controller.imageupdateautomation","msg":"Reconciler error","reconciler group":"image.toolkit.fluxcd.io","reconciler kind":"ImageUpdateAutomation","name":"flux-system","namespace":"flux-system","error":"unable to clone '<url>', error: Failed to authenticate SSH session: Unable to extract public key from private key."}
```

Deleting the flux-system secret and changing to an RSA key (by not passing any argument to `flux bootstrap`) appears to fix the issue.

LocalAreaNitwit:
Exactly the same issue here.
`{"level":"error","ts":"2021-06-17T09:46:35.545Z","logger":"controller-runtime.manager.controller.imageupdateautomation","msg":"Reconciler error","reconciler group":"image.toolkit.fluxcd.io","reconciler kind":"ImageUpdateAutomation","name":"flux-system","namespace":"flux-system","error":"unable to clone 'ssh://git@GITREPO', error: Failed to authenticate SSH session: Unable to extract public key from private key."}`

I will try the work around to confirm it fixes it for us too.

hiddeco:
We _think_ this should have been solved in `v0.13.0` of the controller, available in the latest `flux` release.

bootc:
@hiddeco unfortunately no luck for me:
```
{"level":"error","ts":"2021-06-23T09:29:01.736Z","logger":"controller-runtime.manager.controller.imageupdateautomation","msg":"Reconciler error","reconciler group":"image.toolkit.fluxcd.io","reconciler kind":"ImageUpdateAutomation","name":"flux-kubernetes-gitops","namespace":"flux-system","error":"unable to clone 'ssh://git@git.boo.tc/bootc/flux-kubernetes-gitops.git', error: Failed to authenticate SSH session: Unable to extract public key from private key."}
```

rjhenry:
I saw the same with the latest flux image, but was able to work around it by replacing the secret with a custom one featuring an SSH key that I generated  on a machine. It appears as though the private key wasn't in the standard `BEGIN OPENSSH PRIVATE KEY` format.

pjbgf:
[...]

PR #187, opened 2021-06-22
Use Debian as base image

stefanprodan:
This PR changes the base image for the test, build and controller container images to Debian slim.

Ref: https://github.com/fluxcd/source-controller/pull/391
Ref: https://github.com/fluxcd/source-controller/pull/386

Should fix but needs unit tests to prove it #186


stefanprodan:
multi-arch builds without issues https://github.com/fluxcd/image-automation-controller/actions/runs/960373928

hiddeco (approved):

PR #190, opened 2021-06-23
More debug output

squaremo:
This adds trace-level messages to the log, in particular for

 - which setters are created from policies, and which are invoked on the repo
 - which files are scanned as YAMLs
 - which files are added to the commit, and which are ignored

### DIFF
--- a/controllers/imageupdateautomation_controller.go
+++ b/controllers/imageupdateautomation_controller.go
@@ -65,8 +65,16 @@ import (
 	"github.com/fluxcd/image-automation-controller/pkg/update"
 )
 
-// log level for debug info
+// log level for debug output
 const debug = 1
+
+// log level for trace output; the logging system
+// (fluxcd/pkg/runtime/logging) doesn't presently account for levels
+// more verbose than debug, so lump tracing into
+// --log-level=debug. However, it's useful as self-documentation to
+// keep tracing distinct.
+const trace = 1
+
 const originRemote = "origin"
 
 const defaultMessageTemplate = `Update from image update automation`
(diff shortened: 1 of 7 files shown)


controllers/imageupdateautomation_controller.go
```diff
@@ -229,7 +239,7 @@ func (r *ImageUpdateAutomationReconciler) Reconcile(ctx context.Context, req ctr
 		}
 	}
 
-	log.V(debug).Info("cloned git repository", "gitrepository", originName, "ref", ref, "working", tmp)
+	debuglog.Info("cloned git repository", "gitrepository", originName, "ref", ref, "working", tmp)
```
hiddeco, line 242: Wouldn't it be better to log this before the attempt, so that it automatically adds context to any error that may happen during the attempt?

Reason I think this is a better approach, is because reconcilers generally do 3 things while performing operations for a resource:

1. Logging (to tell what is about to happen, or highlight important information during an action).
2. Events, to make certain things that have happened easily accessible to users and machines for observation.
3. Conditions, to advertise observed state. Most often recorded after a reconciler has run an action that changed state, or when it inspected it.

There is some overlap in 1 and 2, but what distinguishes it for me is that 2 in all scenarios I have seen thus far, it never happens before something has taken place, which means that it would be a better task for logging.

squaremo: Yes it would, and that is a well-reasoned insight :+1:

controllers/imageupdateautomation_controller.go
```diff
@@ -65,8 +65,16 @@ import (
 	"github.com/fluxcd/image-automation-controller/pkg/update"
 )
 
-// log level for debug info
+// log level for debug output
 const debug = 1
+
+// log level for trace output; the logging system
```
hiddeco, line 71: Can you create an issue for this in `fluxcd/pkg`? It will likely be useful for other controllers too now that we are standardizing helpers, and will fit well together with the user requests we have had about "debug notifications".

hiddeco (approved):
The lines that have been added and/or changed look like they adhere to the guidelines I had in mind.

Thank you @squaremo ✨

PR #191, opened 2021-06-24
Graduate to v1beta1

squaremo:
This PR does the following:
    
     - copies the type definitions from v1alpha2 to v1beta1
     - changes the "stored" version to v1beta1
     - gives the CRD a conversion strategy of None, meaning just rewrite the version
     - switches the controller to use v1beta1
     - moves the generated documentation, and API guide, to v1beta1
    
This effectively rebadges the v1alpha2 version of this part of the image API to v1beta1. The v1alpha2 version is left in place; there are no conversion issues, as with v1alpha1->v1alpha2. The CRD specifies that converting between v1alpha2 and v1beta1 just means changing the version (i.e., the schema and semantics are the same).

NB this will work happily without waiting for the new image reflector API to be released, since that also is just a rebadging. For the sake of keeping this tidy, either this PR or a future PR should bump the image reflection API version.


stefanprodan (approved):
LGTM

Thanks @squaremo 🥇

PR #192, opened 2021-06-24
feat: add controller concurrency

makkes:
The controller is now working with 4 concurrent workers by default.
This value is configurable through the `--concurrent` flag.

Signed-off-by: Max Jonas Werner <mail@makk.es>

### DIFF
--- a/controllers/imageupdateautomation_controller.go
+++ b/controllers/imageupdateautomation_controller.go
@@ -47,6 +47,7 @@ import (
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/builder"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller"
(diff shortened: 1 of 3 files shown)


controllers/imageupdateautomation_controller.go
```diff
@@ -558,7 +566,7 @@ func commitChangedManifests(tracelog logr.Logger, repo *gogit.Repository, absRep
 	// modified. There's no circumstance in which we want to commit a
 	// change to a broken symlink: so, detect and skip those.
 	var changed bool
-	for file, _ := range status {
+	for file := range status {
```
squaremo, line 569: Nit: this is an unrelated change, best not to include it in this commit/PR.

squaremo (commented):
I have recorded a nitpick; modulo that, this is good to go, thank you @makkes

makkes:
@squaremo Thanks for the review! I addressed your comment.

stefanprodan (approved):
LGTM 

Thanks @makkes ☕

Issue #180, opened 2021-04-11
images are not fetched


kadzielawa:
Hello,
thank you for your powerful tool, I really like it! great job ! it's big milestone to gitops :) 
I have problem with fluxcd2 because during creating processes which would be automatically after pushing images to ECR repository, HelmRelease is not upgrading containers however ImageUpdateAutomation works great and helm values file is updated on git repository very well.

not upgrading containers - I mean that even in repository the image is in the newest version eg.(sdd9) container is still in wrong version...
do you have any ideas what it's wrong with those files?

flux logs doesn't return nothing special... I see only that every one minute updating working.

something like values.yaml are not valid recognize by kubernetes   


```
Every 2.0s: flux get images all                                                                                                                                                                                                                                                                                            iMac-Jakub.local: Sun Apr 11 09:45:36 2021

NAME                    READY   MESSAGE                         LAST SCAN                       SUSPENDED
imagerepository/podinfo True    successful scan, found 34 tags  2021-04-11T09:44:46+02:00       False

NAME                    READY   MESSAGE                                                                                                 LATEST IMAGE
imagepolicy/podinfo     True    Latest image tag for '247635813899.dkr.ecr.eu-central-1.amazonaws.com/hey-flux' resolved to: sdd9      247635813899.dkr.ecr.eu-central-1.amazonaws.com/hey-flux:sdd9

NAME                                    READY   MESSAGE                                                         LAST RUN                        SUSPENDED
imageupdateautomation/flux-system       True    no updates made; last commit 58dd783 at 2021-04-11T07:35:46Z    2021-04-11T09:44:53+02:00       False


``` 





structure (my-private-repo)
-file1.yml
-file2.yml
-file3.yml
---eks(directory)
-------heyflux(directory)
----------values.yaml
----------Chart.yaml
----------rest chart files

file1.yml
```
apiVersion: image.toolkit.fluxcd.io/v1alpha1
kind: ImageUpdateAutomation
metadata:
  name: flux-system
  namespace: flux-system
spec:
  checkout:
    branch: master
    gitRepositoryRef:
      name: podinfo
  commit:
    authorEmail: fluxcdbot@users.noreply.github.com
    authorName: fluxcdbot
    messageTemplate: '{{range .Updated.Images}}{{println .}}{{end}}'
  interval: 1m0s
  push:
    branch: master
  update:
    path: ./eks/heyflux
    strategy: Setters
```
file2.yml
```
---
[...]

PR #200, opened 2021-07-16
Add a note to SourceRef to address Image FAQ

kingdonb:
This is my first contribution in Image Automation Controller 🎉 I think we can do better than this doc update, but I think this is an important first step to make [this issue easier for users to discover](https://github.com/fluxcd/image-automation-controller/issues/180#issuecomment-877747036) and resolve on their own.

I will open a separate issue for follow-up that I think would make the situation better. (#201)

### DIFF
--- a/api/v1beta1/imageupdateautomation_types.go
+++ b/api/v1beta1/imageupdateautomation_types.go
@@ -27,7 +27,8 @@ const ImageUpdateAutomationKind = "ImageUpdateAutomation"
 // ImageUpdateAutomationSpec defines the desired state of ImageUpdateAutomation
 type ImageUpdateAutomationSpec struct {
 	// SourceRef refers to the resource giving access details
-	// to a git repository.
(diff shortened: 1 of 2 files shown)


api/v1beta1/imageupdateautomation_types.go
```diff
@@ -27,7 +27,8 @@ const ImageUpdateAutomationKind = "ImageUpdateAutomation"
 // ImageUpdateAutomationSpec defines the desired state of ImageUpdateAutomation
 type ImageUpdateAutomationSpec struct {
 	// SourceRef refers to the resource giving access details
-	// to a git repository.
+	// to a git repository. It must be in the same namespace as the
+	// ImageUpdateAutomation.
```
squaremo, line 31: It's not totally obvious what "it" is referring to. Perhaps put the constraint in line with the referent:

```go
// SourceRef refers to the GitRepository object in the same namespace that has access details.
```

Issue #205, opened 2021-07-30
Feature request: configurable branch force pushing


jracollins:
## Use case: 

A workflow I am using currently is to push the image automation updates to a different branch than the one flux is watching. A Github action then opens a PR on any commit to that branch (and the branch is subsequently pruned upon merging). 

This allows users to approve/deploy releases manually, whilst keeping most of the workflow (i.e: committing and pushing tags) automated. 

## Current Issue:

If a branch with image updates is created by flux, and a new commit comes into the main branch that it was branched off, any subsequent updates are blocked because the Git history on the opened branch is out of sync. The "solution" is to rebase the branch and add the image update commit. However this requires a force push to the new branch, which flux currently does not do. 

This will occur when using separate branches for multiple applications image automation updates. 

Eg:
```
/main
/app-a-update 
/app-b-update
```

If using flux to commit on the same branch as it is watching (ie, auto deploy on new image tag) this is a non issue, as there is no rebasing required. 

However if in this example, /app-a-update branch is merged, the commit on /app-b-update needs to be rebased (and then forced pushed), currently this is not possible and fails. The current workaround I am using is to manually delete the branch and wait for flux to recreate.  

## Proposal:

```YAML
apiVersion: image.toolkit.fluxcd.io/v1alpha2
kind: ImageUpdateAutomation
metadata:
  name: my-application
  namespace: flux-system
spec:
  interval: 1m0s
  sourceRef:
    kind: GitRepository
    name: my-application
  update:
    path: ./production
    strategy: Setters
  git:
    checkout:
      ref:
        branch: main
    push:
      branch: my-application-production-update
      force: true # this would be the proposed setting/flag
    commit:
      author:
        email: fluxcdbot@users.noreply.github.com
        name: fluxcdbot
```

This should default to false and remain backward compatible.

Branch protection in GitHub etc can also prevent any unintended rewriting of important branches. 

Initial Discussion in Flux Slack: https://cloud-native.slack.com/archives/CLAJ40HV3/p1627660144049300

squaremo:
> any subsequent updates are blocked because the Git history on the opened branch is out of sync.

I can see why you want branches to be rebased, but I don't understand this statement -- are commits to the PR blocked by GitHub when it is not fast-forwardable? Or is the PR just not mergeable, until it's been rebased.

> The "solution" is to rebase the branch and add the image update commit. However this requires a force push to the new branch, which flux currently does not do.

[...]

PR #206, opened 2021-07-31
Configurable force pushing

jracollins:
Thought I would have a crack at my own issue I raised, so this is a work in progress PR to implement #205 

I'm not very familiar with Go and the test cases are a little fiddly to write (given most of the "actual" work is done by libgit2 and a single character change to the refspec) so will take me a little longer to write those up. I thought I'd open the draft PR so someone else doesn't unnecessarily pick up the issue when I'm almost done. 

In the meantime a couple questions: 
-  It isn't a breaking API change, but should I be implementing for the older api specs (v1alpha1) etc like I have done? (Not sure what the correct process is) 
-  Any issues with/comments on the actual implementation below?


stefanprodan:
> should I be implementing for the older api specs 

No, older API versions are considered readonly. 

PS. Please signoff your commits, see https://github.com/fluxcd/image-automation-controller/pull/206/checks?check_run_id=3208778477

api/v1beta1/git.go
```diff
@@ -87,4 +87,6 @@ type PushSpec struct {
 	// starting point, if it doesn't already exist.
 	// +required
 	Branch string `json:"branch"`
+	// Force specifies wether to force push the branch to origin
+	Force bool `json:"force"`
```
stefanprodan, line 91: This should be:

```go
	// This flag tells the controller to force push the branch to origin. Defaults to false.
	// +optional
	Force bool `json:"suspend,omitempty"`
```

jracollins: I'm assuming you mean just the `omitempty` part and keeping the key as `force`, not `suspend` - just pushed fix

stefanprodan: Yes thank you for changing it.

squaremo:
This is going to be pretty tricky to test. The controller always does a fresh clone and checkout of the push branch; to necessitate a `push --force`, you would need to alter the upstream push branch _after_ the controller has fetched it, but before it tries to push to it. The test will fail, because it doesn't do this interleaving, but expects `--force` to be necessary.

```go
// observe the first commit
					waitForNewHead(localRepo, pushBranch)

					// Create commit on push branch that would conflict with a commit on top of the a branch from the checkout branch
					dummyCommitMessage := "Dummy Commit"
					commitInRepo(cloneLocalRepoURL, pushBranch, dummyCommitMessage, func(tmp string) {})

// origin pushBranch now has a commit "Dummy Commit"

					waitForNewHead(localRepo, pushBranch)

					head, err := localRepo.Reference(plumbing.NewRemoteReferenceName(originRemote, pushBranch), true)
					Expect(err).NotTo(HaveOccurred())
					pushBranchInitialHeadHash := head.String()
					dummyCommit, err := localRepo.CommitObject(head.Hash())
					Expect(err).NotTo(HaveOccurred())
					Expect(dummyCommit.Message).To(Equal(dummyCommitMessage))

// the local clone now (verifiably) has the head at Dummy Commit

					// update the policy and expect another commit in the push branch, deleting local branch to ensure commit is not on top of the dummy commit
					localRepo.DeleteBranch(pushBranch)
					policy.Status.LatestImage = "helloworld:v1.3.0"
					Expect(k8sClient.Status().Update(context.TODO(), policy)).To(Succeed())

// This deletes pushBranch locally, but does nothing to the origin

// When the policy object is updated, the controller will clone the repo from the origin, check out the branch specified, and commit on top of it.

					// Delete local push branch and refetch remote
					localRepo.DeleteBranch(pushBranch)
					waitForNewHead(localRepo, pushBranch)
					head, err = localRepo.Reference(plumbing.NewRemoteReferenceName(originRemote, pushBranch), true)
					Expect(err).NotTo(HaveOccurred())
					fetchedPushBranchLatestHeadHash := head.String()
					pushBranchLatestCommit, err := localRepo.CommitObject(head.Hash())
					Expect(err).NotTo(HaveOccurred())

// this will get the new head of push branch, which is the commit made by the controller (on top of Dummy Commit)

					// Expect push to fail
					// Check with messages for easier debugging
					Expect(dummyCommit.Message).To(Equal((pushBranchLatestCommit.Message)))

// this fails because the head commit is the commit made by the controller

					// Check with hashes for certainty
					Expect(pushBranchInitialHeadHash).To(Equal((fetchedPushBranchLatestHeadHash)))
```

PR #218, opened 2021-09-08
Port to new pkg/runtime helpers

squaremo:
(These agenda from https://github.com/fluxcd/image-reflector-controller/pull/165):

 - [x] Rewrite conditions code to use API in pkg/runtime/conditions (MarkTrue/...False/...Unknown)
 - [x] Use patch helper rather than own patchStatus
 - [x] Exit early without setting status when .spec.suspend (it already did this)
 - [x] Events and metrics helpers

Ref: https://github.com/fluxcd/flux2/issues/1601


stefanprodan (approved):
LGTM

Thanks @squaremo ☕

hiddeco (approved):

PR #221, opened 2021-09-24
Testenv test refactor

darkowlzz:
Introduces testenv based tests.
Refactors all the existing tests to use testenv.

:warning: target branch of the PR is `reconcilers-dev`.


controllers/suite_test.go
```diff
@@ -17,95 +17,63 @@ limitations under the License.
...
-	// want to see what the reconcilers see.
-	k8sClient, err = client.New(cfg, client.Options{Scheme: scheme.Scheme})
-	Expect(err).ToNot(HaveOccurred())
-	Expect(k8sClient).ToNot(BeNil())
+	code := m.Run()
 
-	close(done)
-}, 60)
+	fmt.Println("Stopping the test environment")
+	if err := testEnv.Stop(); err != nil {
```
stefanprodan, line 74: This hangs when used with controller-runtime > 0.10, we need to stop the controller before testenv. Here is an example: https://github.com/fluxcd/kustomize-controller/blob/v1beta2/controllers/suite_test.go#L150-L151

darkowlzz: Hi, I couldn't reproduce that. I used to see similar issues before, when using envtest. But since the refactor, I haven't seen that.
Also, I tried updating to controller-runtime separately, just to reproduce it and instead of the hang issue, I discovered some other failures due to a breaking change in the controller-runtime client, refer: https://github.com/kubernetes-sigs/controller-runtime/pull/1640.
I fixed that and everything works well with CR v0.10.1. I used to be able to run `pgrep` to look for `etcd` and `apiserver` when it used to experience the hanging behavior. I don't get any results anymore. Seems to be stopping everything correctly.
Could it be something to do with the environment? I tried this on a linux environment.

darkowlzz: Update: I create #230 to update controller-runtime separately. While running the tests in that branch, I was able to reproduce the hanging behavior. Although the tests passed, the overall tests failed because the kube-apiserver couldn't be stopped:

```console
  Unexpected error:
      <errors.aggregate | len:1, cap:1>: [
          <*errors.errorString | 0xc0005085d0>{
              s: "timeout waiting for process kube-apiserver to stop",
          },
      ]
      timeout waiting for process kube-apiserver to stop
  occurred
```

Again, I can't reproduce it with these refactored tests.

stefanprodan: Even if it no longer hangs, I think we should close the controller before envtest, otherwise the controller will write dozen error logs to stderr due to failing to connect to the API.

squaremo:
TODO for me to check:

**controllers/update_test.go**

- ImageUpdateAutomation
 - [ ] initialises Git OK
 - commit spec
   - [ ] formats the commit message as in the template
   - [ ] has the commit author as given
 - update path
   - [ ] updates only the deployment in the specified path
 - commit signing
   - [ ] signs the commit with the generated GPG key
 - Using go-git with HTTP
   - with PushSpec
     - [ ] creates and pushes the push branch
     - [ ] pushes another commit to the push branch
     - [ ] still pushes to the push branch after it's merged
   - with Setters
     - [ ] updates to the most recent image
     - [ ] stops updating when suspended
 - Using go-git with SSH
   - with PushSpec
     - [ ] creates and pushes the push branch
     - [ ] pushes another commit to the push branch
     - [ ] still pushes to the push branch after it's merged
   - with Setters
     - [ ] updates to the most recent image
     - [ ] stops updating when suspended
 - Using libgit2 with HTTP
   - with PushSpec
     - [ ] creates and pushes the push branch
     - [ ] pushes another commit to the push branch
     - [ ] still pushes to the push branch after it's merged
   - with Setters
     - [ ] updates to the most recent image
     - [ ] stops updating when suspended
 - Using libgit2 with SSH
   - with PushSpec
     - [ ] creates and pushes the push branch
     - [ ] pushes another commit to the push branch
     - [ ] still pushes to the push branch after it's merged
   - with Setters
     - [ ] updates to the most recent image
     - [ ] stops updating when suspended
 - defaulting
   - defaults .spec.update to {strategy: Setters}

**controllers/git_test.go**

(already in normal form)

**controllers/git_error_test.go**

(already in normal form)

**pkg/test/files_test.go**

 - when no differences
   - [ ] matches when given the same directory
   - [ ] matches when given equivalent directories
 - with differences
   - [ ] finds files in expected from a/ but not in actual from b/
   - [ ] find files in actual a/ that weren't expected from b/
   - [ ] finds files that are different in a/ and b/
 
**pkg/update/filereader_test.go**

 - load YAMLs with ScreeningLocalReader
   - [ ] loads only the YAMLS containing the token

**pkg/update/result_test.go**

 - image ref
   - [ ] gives each component of an image ref
   - [ ] deals with hostnames and digests
 - update results
   - [ ] deduplicates images
   - [ ] collects images by object

**pkg/update/update_test.go**

 - Update image via kyaml/setters2
   - [ ] updates the image marked with the image policy (setter) ref
   - [ ] gives the result of the update

squaremo:
To be honest, I prefer the previous formulation (using ginkgo), which was more readable. To wit: I'm finding it difficult to check whether the same things are tested -- it's fairly obvious in the ginkgo version because the setting up and tearing down is separate to the assertions, and the assertions are separate to each other.

Sometimes the reliance of Ginkgo on closures makes it hard to see what's going on, it's true. But flattening it all out is two steps backward for one step forward, I reckon. If ginkgo absolutely must be removed, then I'd like at least to see some of the same structure preserved, where setup/teardown code is shared.

squaremo:
> I believe it can be refactored in ways to make it better with helpers for common tasks.

I think we can get something better than both ginkgo and "flat" tests, by factoring out some bits (which I found awkward with ginkgo) and using closures judiciously to share setup code.

The particular file I was looking at when I wrote the comment above was controllers/update_test.go, which mixes a lot of what were individual tests together into one long section in `TestImageAutomation_commit_spec`. I think the commit message, path and signing should really have their own tests, and just share a bit of setup.

TestImageAutomation_e2e after that uses a table appropriately, I think -- though it's still very long and could surely have some bits factored out (this was true of the ginkgo code too, it's not a new problem :-).

darkowlzz:
This is ready for another review. I hope this is much easier than before to read. But I'm open for more restructuring and simplification.

The DCO check is failing with:

```
Commit sha: 3917048, Author: Michael Bridgen, Committer: Sunny; Expected "Michael Bridgen mikeb@squaremobius.net", but got "Michael Bridgen mikeb@weave.works".
```

Is it okay if I edit @squaremo 's commit signoff message myself?

squaremo:
> The first part sets up the git server, initializes a repo, creates GitRepository resource for the git repo and creates ImagePolicy. The second part is the actual tests in subtests where depending on the test case, ImageUpdateAutomation object is created and scenario specific operations are performed.

It looks like this scheme is working well in controllers/update_test.go. The first few tests do exactly the same setup (I imagine this is why you mentioned more factoring-out!); perhaps you could use a callback:

```go
func TestSignedCommit(t *testing.T) {
  testWithRepoAndImagePolicy(NewWithT(t), func(tmp, localRepo) {
    policyKey := types.NamespacedName{
      Name:      imagePolicyName,
      Namespace: namespace,
    }
    commitInRepo(g, repoURL, branch, "Install setter marker", func(tmp string) {
      g.Expect(replaceMarker(tmp, policyKey)).To(Succeed())
    })
    // ...
  })
}
```

(I've glossed over a lot of detail there of course, it's possible that the callback would need too many arguments)

darkowlzz:
Consolidated the common test setup code into `testWithRepoAndImagePolicy()` which also accepts a `testFunc` callback to run the test assertions after the setup.
It did result in passing a lot of arguments, but I guess it's okay the way it is now, manageable.

Ready for a review.

squaremo (approved):
I'm satisfied that the rewritten suite contains the same tests (thanks for making that pretty easy to check!). The new structure, with subtests sharing setup and common funcs factored out, is at least as good as the structure when using ginkgo I think, and has less magic and fewer lines :+1:

I made a PR with a couple of suggestions, but I don't consider them prerequisites to merging this PR.

Thanks for persisting with this work, Sunny 🍍

controllers/testdata/appconfig-setters-expected/deploy.yaml
```diff
@@ -7,4 +7,4 @@ spec:
     spec:
       containers:
         - name: hello
-          image: helloworld:1.0.1 # SETTER_SITE
+          image: helloworld:v1.3.1 # SETTER_SITE
```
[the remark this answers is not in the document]
darkowlzz: Yes, it was deliberate or incidentally add to adapt to the new structure because the subtests shared the same image policy.
I just tried moving the image policy from being common to all the subtests, to be created per subtests. That helps avoid changing the above testdata and removes the dependence on the ordering of the tests. Much better now.
Thanks for emphasizing this.

pkg/update/result_test.go
```diff
@@ -18,79 +19,76 @@ func mustRef(ref string) imageRef {
 	return imageRef{r, types.NamespacedName{}}
 }
 
-var _ = Describe("image ref", func() {
-	It("gives each component of an image ref", func() {
+func TestMustRef(t *testing.T) {
```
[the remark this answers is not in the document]
darkowlzz: Changed.

darkowlzz:
Rebased, merged the readability improvements by @squaremo and updated as per the last few suggestions to make the e2e tests more independent.

Issue #160, opened 2021-05-04
Support commit message template functions


Dev25:
The [current implementation](https://github.com/fluxcd/image-automation-controller/blob/main/controllers/imageupdateautomation_controller.go#L276-L288) of a commit message template relies on Go's `text/template` with no extra functions, it would be great to be able to use something like the sprig library which provides a bunch of functions (what Helm uses) or alternatively implement a small amount of functions depending on user demand.

http://masterminds.github.io/sprig/

This then allows users to have more complex logic when generating commits e.g. the following only works when there is 1 deployment, if i have multiple deployments following the same image policy then i need to count the unique strings in `.Updated.Images` which is not possible with the default `text/template` functions.
```go
        {{ if eq (len .Updated.Images) 1 -}}
        [skip ci] Auto Release of {{ index .Updated.Images 0 }}
        {{ else -}}
        [skip ci] Auto Release
        {{ end }}
```

souleb:
can work on that.

squaremo:
This seems like a decent idea and a reasonable request :-) My reservations are:

 - if sprig drags in a lot of dependencies, that makes upkeep more difficult (e.g., more chance for CVEs to apply)
 - we don't really want any functions that get access to the network or OS; if those can be disabled, then :+1:.

Another specific, tricky case is functions that access a source of randomness; I would err on the side of not including them.

squaremo:
Re accessing OS and randomness: https://pkg.go.dev/github.com/Masterminds/sprig/v3#HermeticTxtFuncMap

stefanprodan:
Implemented in #223

PR #223, opened 2021-09-29
Add support for the sprig functions library

souleb:
Adresses #160

All functions are not supported, The following are discarded:
```
	// Date functions
	"date",
	"date_in_zone",
	"date_modify",
	"now",
	"htmlDate",
	"htmlDateInZone",
	"dateInZone",
	"dateModify",

	// Strings
	"randAlphaNum",
	"randAlpha",
	"randAscii",
	"randNumeric",
	"uuidv4",

	// OS
	"env",
	"expandenv",

	// Network
	"getHostByName",
```
Commit message templating can use those functions

Signed-off-by: Soule BA <bah.soule@gmail.com>


stefanprodan (commented):
@souleb please add a section here https://github.com/fluxcd/image-automation-controller/blob/main/docs/spec/v1beta1/imageupdateautomations.md#commit-message-template-data about spring functions and provide an example.

controllers/imageupdateautomation_controller.go
```diff
@@ -793,3 +787,24 @@ func (r *ImageUpdateAutomationReconciler) recordSuspension(ctx context.Context,
...
 }
+
+// TemplateMsg renders a msg template, returning the message or an error.
+func TemplateMsg(messageTemplate string, templateValues *TemplateData) (string, error) {
+	if messageTemplate == "" {
+		messageTemplate = defaultMessageTemplate
+	}
+
+	// All functions taht are not guaranteed to evaluate to the same result for given input, because they
+	// refer to the environment or global state are excluded.
```
squaremo, line 798:
```suggestion
	// Includes only functions that are guaranteed to always evaluate to the same result for given input. This removes the possibility of accidentally relying on where or when the template runs.
```

controllers/imageupdateautomation_controller.go
```diff
@@ -793,3 +787,24 @@ func (r *ImageUpdateAutomationReconciler) recordSuspension(ctx context.Context,
 		r.MetricsRecorder.RecordSuspend(*objRef, auto.Spec.Suspend)
 	}
 }
+
+// TemplateMsg renders a msg template, returning the message or an error.
+func TemplateMsg(messageTemplate string, templateValues *TemplateData) (string, error) {
```
squaremo, line 792: Does this need to be exported? I think it's better not to, by default. (You can accuse me of hypocrisy, I've probably exported random funcs elsewhere in the controller code ...)

squaremo (commented):
Neatly done, thank you @souleb. I've made a couple of suggestions; see what you think, and if you're happier leaving things as they are, respond with a comment.

squaremo:
Additionally: first contribution high-five ✋ !

controllers/imageupdateautomation_controller.go
souleb: Thanks! Definitely shouldn't be exported

squaremo:
This will make lots of users happy, thank you @souleb!

PR #224, opened 2021-09-29
Fix nil-dereference in controller

DavidKorczynski:
Fixes a nil-deref when the following conditions are all true:

- `gitSpec.Checkout` is nil
- `origin.Spec.Reference` is nil
- `gitSpec.push` is nil

Signed-off-by: David Korczynski <david@adalogics.com>

### DIFF
--- a/controllers/imageupdateautomation_controller.go
+++ b/controllers/imageupdateautomation_controller.go
@@ -211,8 +211,8 @@ func (r *ImageUpdateAutomationReconciler) Reconcile(ctx context.Context, req ctr
 		// Here's where it gets constrained. If there's no push branch
 		// given, then the checkout ref must include a branch, and
 		// that can be used.
-		if ref.Branch == "" {
-			failWithError(fmt.Errorf("Push branch not given explicitly, and cannot be inferred from .spec.git.checkout.ref or GitRepository .spec.ref"))
+		if ref == nil || ref.Branch == "" {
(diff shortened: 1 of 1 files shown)


DavidKorczynski:
Discussed this with @squaremo internally

controllers/imageupdateautomation_controller.go
```diff
@@ -216,6 +216,8 @@ func (r *ImageUpdateAutomationReconciler) Reconcile(ctx context.Context, req ctr
 		}
 		pushBranch = ref.Branch
 		tracelog.Info("using push branch from $ref.branch", "branch", pushBranch)
+	} else {
+		failWithError(fmt.Errorf("Invalid reference. Please either specify a push branch, .spec.git.checkout.ref or spec.ref"))
```
squaremo, line 220: I think this is the same failure case as `ref.Branch == ""`, so it would be better to put this branch in the same place, to fail with the same error.

DavidKorczynski: Agreed - please do see my comment below on adding a return statement too

DavidKorczynski:
This PR also fixes another issue which is it adds a `return` on the `failWithError` - this didn't occur in the specified place which cause the reconciler to continue execution after the call to `failWithError`

squaremo (approved):
Well spotted on both counts, thank you @DavidKorczynski 🍎

PR #230, opened 2021-10-01
Update controller-runtime to v0.10.1

darkowlzz:
Update controller-runtime to v0.10.0 and update the tests to be
compatible with the breaking changes.

In controller-runtime v0.10.0, the client is updated to clean any stale
data in the target object when performing any operation. This results in
test failure for the code that constructs an object with both spec and
status, and creates the object and updates status with the same
object. The fix is to set the status separately on the object before
updating it.
Refer: https://github.com/kubernetes-sigs/controller-runtime/pull/1640


stefanprodan (approved):
LGTM

Thanks @darkowlzz

PR #251, opened 2021-11-01
Respect PKG_CONFIG_PATH from the environment

squaremo:
If brew (or some other installation method) has put openssl.pc in a non-traditional place, you need to add that place to `PKG_CONFIG_PATH`; but, the Makefile effectively resets it when calling anything that uses it.

So:
 - include any existing $PKG_CONFIG_PATH in invocations
 - document when and how to set PKG_CONFIG_PATH

---

Note: this doesn't appear to be necessary for source-controller. Hidde and I think this may be because git2go is a direct dependency there, whereas it's required via source-controller/pkg here -- why that matters is not clear :-/

### DIFF
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -27,6 +27,40 @@ $ brew install cmake openssl@1.1 libssh2 pkg-config
 $ LIBGIT2_FORCE=1 make libgit2
 ```
 
+#### openssl and pkg-config
+
+You may see this message when trying to run a build:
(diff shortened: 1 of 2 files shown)


hiddeco (approved):
Let's just hope the ducktape holds longer than it takes for git2go to have bug free Go transport

WOMM (Works on Michael's Machine) ✅

darkowlzz (approved):
LGTM!

PR #252, opened 2021-11-01
Pass context to libgit2.RemoteCallbacks

darkowlzz:
Pass cancellable context to libgit2.RemoteCallbacks to be able to cancel
the remote operations when the context is cancelled.
For git clone, the context is created with the timeout specified in the target
GitRepository.

Based on the source-controller changes in https://github.com/fluxcd/source-controller/pull/477.


hiddeco (approved):
LGTM, thank you @darkowlzz :bow:

PR #253, opened 2021-11-02
config/crd: Update the generated CRD spec

darkowlzz:
source-controller v0.17.0 updated the `GitRepositoryRef` definition which
updates `GitCheckoutSpec.Reference` in IAC APIs.


darkowlzz:
Closing this and adding the same changes as part of #252 .

Issue #275, opened 2021-12-07
Set default observedGeneration to -1


makkes:
This is a follow-up for the tracking issue fluxcd/flux2#2007.

The default observedGeneration has to be set on the following CRDs:

* ImageUpdateAutomation

sbernheim:
@makkes - This seems pretty straightforward, and I have a commit ready to resolve it based on @tomhuang12 's previous PR. fluxcd/image-reflector-controller#189 (Thanks, Tom!)

I'd like to submit a PR for this Issue during @scottrigby 's Controller Collab session tomorrow to demonstrate the contribution process for first-timers, if that's OK.  I'll submit the PR tomorrow afternoon regardless if we don't get to it during the class.

You can ping me on the Weaveworks or CNCF Slack to let me know if you would prefer that I not this Issue as a `good-first-issue` classroom example.

stefanprodan:
Fixed in #277

PR #277, opened 2021-12-10
Set default observedGeneration to -1

sbernheim:
Sets a default value of -1 for the `observedGeneration` field of the ImageUpdateAutomations type `status.observedGeneration` attribute. This ensures that tools like `kstatus` do not consider the resource to be in a Ready state prematurely because the `generation` and `observedGeneration` attributes are briefly initialized with `0` values.

- https://github.com/fluxcd/flux2/issues/2007
- https://github.com/fluxcd/image-automation-controller/issues/275

Credits to @tomhuang12 for his prior work on fluxcd/image-reflector-controller#189

### DIFF
--- a/api/v1beta1/imageupdateautomation_types.go
+++ b/api/v1beta1/imageupdateautomation_types.go
@@ -129,7 +129,8 @@ type ImageUpdateAutomation struct {
 	metav1.TypeMeta   `json:",inline"`
 	metav1.ObjectMeta `json:"metadata,omitempty"`
 
-	Spec   ImageUpdateAutomationSpec   `json:"spec,omitempty"`
+	Spec ImageUpdateAutomationSpec `json:"spec,omitempty"`
+	// +kubebuilder:default={"observedGeneration":-1}
 	Status ImageUpdateAutomationStatus `json:"status,omitempty"`
 }
 
--- a/config/crd/bases/image.toolkit.fluxcd.io_imageupdateautomations.yaml
+++ b/config/crd/bases/image.toolkit.fluxcd.io_imageupdateautomations.yaml
@@ -690,6 +690,8 @@ spec:
             - sourceRef
             type: object
           status:
+            default:
+              observedGeneration: -1
             description: ImageUpdateAutomationStatus defines the observed state of
               ImageUpdateAutomation
             properties:


sbernheim:
Attempting to correct for the failed check now.

stefanprodan:
@sbernheim you shouldn’t edit edit the manifests manually as the generated from code. Please run `make manifests`.

sbernheim:
l had to do a bunch of things to get the tests to run.

The tests don't seem to want to work without Docker which I was trying to replace with `containerd` and `podman`.  In the end I gave up on figuring out how to make that work and reinstalled Docker for Desktop on my MacBook again.

I also had to add an `.envrc` file with this env var export:
```
export PKG_CONFIG_PATH="/usr/local/Cellar/openssl@1.1/1.1.1m/lib/pkgconfig"
```
That directory is where `brew install openssl@1.1` put the `openssl.pc` file that needs to be linked for the tests.

Now of course the tests are failing but with a different error message that the GitHub checks which indicates that the default `observedGeneration` value is supposed to be an integer and not a string.  The `make manifests` command generates the CRD file with the default value set to a string `"-1"` instead of an integer `-1` and I'm not sure how to fix that.

All the long string values in the generated CRD's `description` fields are also wrapped at 90 characters, which seems like it'd be a linter setting.

However, the same tests fail on my local and the generated CRD has the same problem when I run it against the `main` branch for the `image-reflection-controller` project.  So I have some sort of version conflict between my environment and yours - probably the wrong `kubebuilder` version installed on my local.  I installed the latest today while setting all this up, and more likely need to drop back to an older version.

I'll dig a bit more and try to resolve these issues over the next week or so.

sbernheim:
Found an open issue for the line-wrapping problem, which has a suggested mitigation in the most recent comment https://github.com/kubernetes-sigs/controller-tools/issues/514

Putting the link here so I don't forget where I saw that later when I have time to try again.

sbernheim:
Pinning `controller-gen` to v0.5.0 seems to have resolved the problem of the default value being set as a string in the generated CRD.
```
go get sigs.k8s.io/controller-tools/cmd/controller-gen@v0.5.0
```

It's still wrapping all the long text lines at 90 chars, though.  I pushed what I have to see if the tests pass, but obviously I'd rather not change all the `description` fields like this.  I'll have to circle back and figure out how to fix that another day.

makkes:
@sbernheim I created a commit on my own fork that should fix CI: https://github.com/makkes/image-automation-controller/commit/28fc2cb5921b0f2485512c70ba844e649fcf64d6

Feel free to cherry-pick it into this PR's branch.

sbernheim:
@makkes - Thanks for that!  I cherry-picked your PR into my branch, but the "dirty tree" check still failed!

It appears that @relu merged [a commit](https://github.com/fluxcd/image-automation-controller/commit/24594ebf20bc83a06973b1a03293853b08cf5185) to main while updating the Flux `pkg` components that wrapped text lines at 90 characters that match the ones I'm generating on my local.

Took me a while to figure out why that last round of testing was failing.  At first I thought I must be reading the diff backwards and had somehow dropped your commit again after merging it to my branch.  I figured out what had happened once I took a little dive into the blame history for the generated CRD file.

Hopefully the tests will succeed this time and we can finally put this one to bed.  All's well that ends well AFAIC!

go.sum
```diff
@@ -771,6 +785,12 @@ github.com/onsi/ginkgo v1.10.3/go.mod h1:lLunBs/Ym6LB5Z9jYTR76FiuTmxDTDusOGeTQH+
 github.com/onsi/ginkgo v1.11.0/go.mod h1:lLunBs/Ym6LB5Z9jYTR76FiuTmxDTDusOGeTQH+WWjE=
 github.com/onsi/ginkgo v1.12.1/go.mod h1:zj2OWP4+oCPe1qIXoGWkgMRwljMUYCdkwsT2108oapk=
 github.com/onsi/ginkgo v1.14.0/go.mod h1:iSB4RoI2tjJc9BBv4NKIKWKya62Rps+oPG/Lv9klQyY=
+<<<<<<< HEAD
```
stefanprodan, line 788: You need to resolve all conflicts here, use an editor for Git maybe that’s easier

relu: Running `go mod tidy` should do it.

relu (changes requested):

go.sum
sbernheim: If only there was an eye-roll emoji on GitHub...

I actually DID EXACTLY THIS while merging the fix @makkes gave me for the CRD linebreaks but like an idiot I cherry-pick-merged them into the same commit.  Which of course, I then dropped later when I realized I had to undo the line-break changes, and forgot about before pushing again.

The comedy (of errors) never stops!  I'll be here all week.  Tip your waitress.

relu (approved):
LGTM, thank you @sbernheim! :heavy_check_mark:

sbernheim:
🎉 !!!

PR #278, opened 2021-12-13
Fix potentially broken support for macOS

aryan9600:


stefanprodan (approved):
LGTM

Thanks @aryan9600

Issue #246, opened 2021-10-27
Possible nil-deref in image-automation controller


dholbach:
From Ada Logics:

> Consider the following code:

https://github.com/fluxcd/image-automation-controller/blob/674f833d982628831c101a919c0154c64c455db3/controllers/imageupdateautomation_controller.go#L310-L320

> When the `lastTime.Format` is called it is not certain that `lastTime` is not a nil-pointer. As such, similar to a check on `lastCommit` there should be a check on whether `lastTime` is nil.

Nalum:
I'd like to take up this issue if that's good with you

Nalum:
At the moment there is no error message returned if `lastCommit` is an empty string, is this okay to stay this way or should there be an `else failWithError` at this point?

Nalum:
If `lastCommit` is not an empty string but `lastTime` is `nil` do you still want `lastCommit` in `statusMessage`?

I feel like it's unlikely to happen that you would have one without the other but it is possible.

hiddeco:
I feel the presence of both is required for the observation to be correct.

PR #283, opened 2021-12-17
Check for a nil lastTime before use

Nalum:
Fix: #246 

Avoiding a nil-deref by checking the `lastTime` variable before using it.

### DIFF
--- a/controllers/imageupdateautomation_controller.go
+++ b/controllers/imageupdateautomation_controller.go
@@ -320,14 +320,15 @@ func (r *ImageUpdateAutomationReconciler) Reconcile(ctx context.Context, req ctr
 	}
 
 	if rev, err := commitChangedManifests(tracelog, repo, tmp, signingEntity, author, message); err != nil {
-		if err == errNoChanges {
-			debuglog.Info("no changes made in working directory; no commit")
-			statusMessage = "no updates made"
-			if lastCommit, lastTime := auto.Status.LastPushCommit, auto.Status.LastPushTime; lastCommit != "" {
-				statusMessage = fmt.Sprintf("%s; last commit %s at %s", statusMessage, lastCommit[:7], lastTime.Format(time.RFC3339))
-			}
-		} else {
+		if err != errNoChanges {
 			return failWithError(err)
+		}
+
+		debuglog.Info("no changes made in working directory; no commit")
+		statusMessage = "no updates made"
+
+		if auto.Status.LastPushTime != nil && len(auto.Status.LastPushCommit) >= 7 {
+			statusMessage = fmt.Sprintf("%s; last commit %s at %s", statusMessage, auto.Status.LastPushCommit[:7], auto.Status.LastPushTime.Format(time.RFC3339))
 		}
 	} else {
 		// Use the git operations timeout for the repo.


Nalum:
Not too sure about how to test this, happy to get some pointers.

controllers/imageupdateautomation_controller.go
```diff
@@ -319,7 +319,7 @@ func (r *ImageUpdateAutomationReconciler) Reconcile(ctx context.Context, req ctr
 		if err == errNoChanges {
 			debuglog.Info("no changes made in working directory; no commit")
 			statusMessage = "no updates made"
-			if lastCommit, lastTime := auto.Status.LastPushCommit, auto.Status.LastPushTime; lastCommit != "" {
+			if lastCommit, lastTime := auto.Status.LastPushCommit, auto.Status.LastPushTime; lastCommit != "" && lastTime != nil {
 				statusMessage = fmt.Sprintf("%s; last commit %s at %s", statusMessage, lastCommit[:7], lastTime.Format(time.RFC3339))
```
pjbgf, line 323: `lastCommit[:7]` will panic if lastCommit string is not as long as it assumes it is (e.g. at least 7 characters long). Therefore the if has to confirm that assumption.

pjbgf (changes requested):
> Not too sure about how to test this, happy to get some pointers.

I can't see an easy way to test it, apart from extracting the if-else block of code into its own function and created a unit test just for it.

controllers/imageupdateautomation_controller.go
Nalum: Good catch, I hadn't picked up on this.

Nalum:
> I can't see an easy way to test it, apart from extracting the if-else block of code into its own function and creating a unit test just for it.

I can look at doing this but is seems a little overkill to me.

controllers/imageupdateautomation_controller.go
```diff
@@ -316,15 +316,24 @@ func (r *ImageUpdateAutomationReconciler) Reconcile(ctx context.Context, req ctr
...
+			lpcLen := len(auto.Status.LastPushCommit)
+			message := `%s; last commit "%s" at %s`
+
+			if lpcLen > 7 {
+				lpcLen = 7
+				message = `%s; last commit "%s..." at %s`
+			}
+
+			statusMessage = fmt.Sprintf(message, statusMessage, auto.Status.LastPushCommit[:lpcLen], auto.Status.LastPushTime.Format(time.RFC3339))
+		}
```
pjbgf, line 336: nit: could be simplified for clarity

```suggestion
		if auto.Status.LastPushTime != nil && len(auto.Status.LastPushCommit) >= 7 {
			statusMessage = fmt.Sprintf("%s; last commit %s at %s", statusMessage, auto.Status.LastPushCommit[:7], auto.Status.LastPushTime.Format(time.RFC3339))
		}
```

pjbgf (commented):
> I can look at doing this but is seems a little overkill to me.

agreed, besides the ongoing refactoring may help on that front too.

on a side point, I added a nit comment about simplifying the if logic for clarity.

controllers/imageupdateautomation_controller.go
Nalum: I think you've missed a bit of what the new code is doing. The internal `if` on the length of the commit message is only trimming it if it's `>7`, the suggested code will only log the message if the commit message is `>7`. I can add a comment around this given that it's not clear what's happening, if that would help.

pjbgf: That was intentional, as 7 characters is the short version of the hash used to represent commits. Decreasing the length here will increase the likelihood of collisions. The idea on the suggestion was to also maintain the current behaviour.

As this field represents the SHA1 commits, it should contain either the full 40 characters or 7 characters for the short version. I would assume something went wrong earlier in the process for a non-empty string with length of anything less than 7.

GitHub requires commit hashes to be at least 4 characters long (in some of their GET APIs). But I am not convinced we benefit by decreasing this.

Nalum: Ah, okay my mistake. I thought this was the commit message not the commit hash. If this is the hash then I'll make that change :+1:

pjbgf (approved):
Nice one @Nalum, thank you for your contribution. :bow: 

LGTM

Nalum:
Thanks for your help and patience @pjbgf  :+1:

stefanprodan:
@Nalum please squash all commits into a single one.

Issue #245, opened 2021-10-27
Unnecessary conditions in the code


dholbach:
From Ada Logics

> In the following code, the `auto.Spec.Update` is not needed, and if indeed `auto.Spec.Update` was nil then the first conditional statement would be panic due to a nil-dereference.

https://github.com/fluxcd/image-automation-controller/blob/7ec4e6150ee0b6f4770a3265161cd98108c06603/controllers/imageupdateautomation_controller.go#L253-L264

> Recommendation
> Remove unnecessary code.

Nalum:
I'd like to take up this issue if that's good with you

PR #284, opened 2021-12-20
Move Path check into switch case

Nalum:
Fix: #245

### DIFF
--- a/controllers/imageupdateautomation_controller.go
+++ b/controllers/imageupdateautomation_controller.go
@@ -249,16 +249,6 @@ func (r *ImageUpdateAutomationReconciler) Reconcile(ctx context.Context, req ctr
 		}
 	}
 
-	manifestsPath := tmp
-	if auto.Spec.Update.Path != "" {
-		tracelog.Info("adjusting update path according to .spec.update.path", "base", tmp, "spec-path", auto.Spec.Update.Path)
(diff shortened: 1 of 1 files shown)


controllers/imageupdateautomation_controller.go
```diff
@@ -269,6 +259,16 @@ func (r *ImageUpdateAutomationReconciler) Reconcile(ctx context.Context, req ctr
...
 		}
 
+		manifestsPath := tmp
+		if auto.Spec.Update.Path != "" {
+			tracelog.Info("adjusting update path according to .spec.update.path", "base", tmp, "spec-path", auto.Spec.Update.Path)
+			if p, err := securejoin.SecureJoin(tmp, auto.Spec.Update.Path); err != nil {
+				return failWithError(err)
+			} else {
+				manifestsPath = p
+			}
```
pjbgf, line 269: nit: `else` is redundant when followed by a `return`.
```suggestion
			p, err := securejoin.SecureJoin(tmp, auto.Spec.Update.Path);
			if err != nil {
				return failWithError(err)
			}
			manifestsPath = p
```

Nalum: Thanks @pjbgf :+1:

pjbgf (approved):
LGTM

stefanprodan (approved):
LGTM

Thanks @Nalum 🏅

PR #287, opened 2022-01-05
Log the error when removing the working dir fails

squaremo:
os.RemoveAll can return an error; even if we're exiting the procedure
at the time, it's worth knowing when it's failing, because typically
the working directory is a memory-backed volume and you can run out
quite easily.


stefanprodan (approved):
LGTM

Thanks @squaremo

PR #290, opened 2022-01-10
Fix Indentation of Makefile

jonathan-innis:
Makefile format is succeeding on MacOS but is causing issues with builds linux-based machines with

```
Makefile:136: *** recipe commences before first target.  Stop.
```


stefanprodan (approved):
LGTM 

Thanks @jonathan-innis

PR #292, opened 2022-01-12
Fix makefile envtest and controller-gen usage

aryan9600:
Refactor logic to install helper tools into one function in the
Makefile. Add support for envtest to help install tools like kubectl,
etcd which helps users run tests more conveniently.

Ref: https://github.com/fluxcd/flux2/issues/2273

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -20,12 +20,6 @@ jobs:
         key: ${{ runner.os }}-go-${{ hashFiles('go.sum') }}
         restore-keys: |
           ${{ runner.os }}-go-
-    - name: Set up kubebuilder
-      uses: fluxcd/pkg/actions/kubebuilder@main
-    - name: Setup envtest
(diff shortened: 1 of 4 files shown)


.github/workflows/build.yaml
```diff
@@ -20,16 +20,8 @@ jobs:
...
-    - name: Set up kubebuilder
-      uses: fluxcd/pkg/actions/kubebuilder@main
-    - name: Setup envtest
-      uses: fluxcd/pkg/actions/envtest@main
-      with:
-        version: "1.19.2"
     - name: Run tests
       uses: ./.github/actions/run-tests
-      env:
-        GOPATH: /github/home/go
```
stefanprodan, line 32: This disables the Go module cache, please undo this change.

stefanprodan (changes requested):

stefanprodan (approved):

PR #295, opened 2022-01-19
Drop capabilities, enable seccomp and enforce runAsNonRoot

aryan9600:
Further restricts the SecurityContext that the controller runs under, by enabling the default seccomp profile and dropping all linux capabilities.
This was set at container-level to ensure backwards compatibility with use cases in which sidecars are injected into the source-controller pod
without setting less restrictive settings.
Add a uid and gid for the container to enforce runAsNonRoot and ensure
the use of non root users.

Ref: fluxcd/flux2#2014

BREAKING CHANGES:
1) The use of new seccomp API requires Kubernetes 1.19.
2) the controller container is now executed under 65534:65534 (userid:groupid).
   This change may break deployments that hard-coded the user name 'controller' in their PodSecurityPolicy.

Co-authored-by: Paulo Gomes <paulo.gomes@weave.works>

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -69,10 +69,6 @@ FROM prepare-${BASE_VARIANT} as build
 
 FROM debian:${BASE_VARIANT}-slim as controller
 
-# Configure user
-RUN groupadd controller && \
-    useradd --gid controller --shell /bin/sh --create-home controller
-
 # Copy libgit2
 COPY --from=build /libgit2/lib/ /usr/local/lib/
 RUN ldconfig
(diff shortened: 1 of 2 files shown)


aryan9600:
This was tested on Kubernetes 1.23.1 with restrict "pod security" (below) and it worked as expected.

```
apiVersion: v1
kind: Namespace
metadata:
  name: flux-system
  labels:
    pod-security.kubernetes.io/enforce: restricted
    pod-security.kubernetes.io/enforce-version: latest
    pod-security.kubernetes.io/warn: restricted
    pod-security.kubernetes.io/warn-version: latest
```

stefanprodan (approved):

Issue #294, opened 2022-01-19
Support cross-namespace sourceRef in ImageUpdateAutomation


squaremo:
For some deployments of Flux, the GitRepository object giving access to a git repo is owned by one team, and the definitions for image updates (ImageRepository, ImagePolicy, and ImageUpdateAutomation) are owned by other teams. There is a strong preference in those cases for keeping the GitRepository in one namespace, and letting teams define automation in their own namespaces -- but to be able to do that, the ImageUpdateAutomation object would need to be able to refer to a GitRepository object in another namespace.

Similar work was done in https://github.com/fluxcd/image-reflector-controller/pull/162, and for more background on similar use cases, see https://github.com/fluxcd/image-automation-controller/issues/85.

(I am not going to recapitulate the pros and cons of breaking namespace isolation here -- that ship has sailed <sup>[meaning](https://idioms.thefreedictionary.com/ship+has+sailed)</sup>.)

aryan9600:
Looking at fluxcd/image-reflector-controller#162, do we want to add similar ACL support or let `ImageUpdateAutomation` objects in any namespace have access to a `GitRepository `?

squaremo:
> do we want to add similar ACL support or let `ImageUpdateAutomation` objects in any namespace have access to a `GitRepository`?

Good question. I would feel more comfortable waiting for `accessFrom` to be implemented in source controller, then respecting it here.

stefanprodan:
> Good question. I would feel more comfortable waiting for accessFrom to be implemented in source controller, then respecting it here.

ACLs are client-side, as described in https://github.com/fluxcd/flux2/pull/2092

stefanprodan:
@squaremo according to https://github.com/fluxcd/flux2/pull/2093 this controller should accept cross-namespace refs as KC and HC do now. When that RFC is merged, only then we can add a flag to enforce ACLs.

Issue #201, opened 2021-07-16
Make "Cross Namespace Reference" error more difficult to trigger or easier to diagnose


kingdonb:
People seem to be wanting to put the automation resources in the same namespace as the Kustomization deploys into. Unfortunately this cannot work, since ImageUpdateAutomation defines SourceRef as a [`kind` and `name` only](https://fluxcd.io/docs/components/image/automation-api/#image.toolkit.fluxcd.io/v1beta1.SourceReference) with no `namespace` field.

This is true for reasons that are well understood, (secrets cannot be read across namespace boundaries, and this will not be possible to change at least in the current incarnation of Flux's model for security that relies on Service Accounts, which are confined to within a namespace.)

But this information still seems to be lacking in support from the docs. The API docs themselves should clarify this minimally in the documentation of `SourceRef`, which I think is addressed in #200.

It still seems likely that by the time someone finds this note in the API docs, they will have already struggled for some time. I think someone needs to look at this issue holistically and figure out if there is more that we could be doing at the point of failure to help the user figure out what they've done wrong. (This will be tricky in some cases especially where there is no actual reconciler actively invoked by the certain failure modes. For example, kyaml tag referring to a policy that does not exist... no automation scanner will currently find this error, and thus no events will be emitted.)

This error is especially common for Helm Controller users, since the HelmRelease resources also define a field `SourceRef`, but with a different `CrossNamespaceObjectReference` type which *actually can* successfully reach across namespaces. This one doesn't have a need for reading directly from any secrets, so it works. I think this difference is very surprising and still seems quite hard to discover.

We seem to get this question from a different user in #flux Slack at least twice a week, and most but not all of them seem to be using Helm Controller. There are a few different ways this failure happens, and I think some of them can be stopped by emitting an error, while some other modes of failing here might already emit errors:

* the `ImageUpdateAutomation.spec.sourceRef` refers to a resource in a different namespace, with a namespace field
* the `sourceRef` refers to a resource which is in a different namespace, without including a reference to the namespace
* the kyaml setter tag references an imagepolicy by name and namespace, but it is not referenced in the correct namespace

The first error should be stopped by existing validation, although I'm not 100% sure about that. There is no namespace field, and if you provide one there, Kubernetes should emit a validation error preventing the resource from being created.

The second seems possible to catch, by emitting an error event from ImageUpdateAutomation signaling that the GitRepository referenced by `sourceRef` could not be found. I think this is something we don't currently emit errors about.

[...]

PR #299, opened 2022-01-21
Add support for cross-namespace sourceRef in ImageUpdateAutomation

aryan9600:
ImageUpdateAutomation objects can now refer to GitRepository objects in other
namespaces. Implemented by switching sourceRef from a SourceReference to a
dependency.CrossNamespaceDependencyReference.

Fix: #294 
Fix: #201
Fix: #180


api/v1beta1/imageupdateautomation_types.go
```diff
@@ -29,7 +30,7 @@ type ImageUpdateAutomationSpec struct {
 	// SourceRef refers to the resource giving access details
 	// to a git repository.
 	// +required
-	SourceRef SourceReference `json:"sourceRef"`
+	SourceRef dependency.CrossNamespaceDependencyReference `json:"sourceRef"`
```
stefanprodan, line 33: You need to create a type like this one https://github.com/fluxcd/kustomize-controller/blob/main/api/v1beta2/reference_types.go with `+kubebuilder:validation:Enum=GitRepository`

stefanprodan (changes requested):

api/v1beta1/imageupdateautomation_types.go
stefanprodan: Using `CrossNamespaceDependencyReference` breaks the API, but because this controller lacks any e2e tests such changes could make it into release. We should get the IRC and IAC in line with all the others controllers and write proper e2e tests.

api/v1beta1/reference.go
```diff
@@ -16,20 +13,32 @@ limitations under the License.
...
+// typed Kubernetes resource object at cluster level.
+type CrossNamespaceSourceReference struct {
+	// API version of the referent.
 	// +optional
 	APIVersion string `json:"apiVersion,omitempty"`
 
-	// Kind of the referent
+	// Kind of the referent.
 	// +kubebuilder:validation:Enum=GitRepository
-	// +kubebuilder:default=GitRepository
```
stefanprodan, line 28: Please restore the default, removing it implies a breaking change.

stefanprodan (changes requested):

stefanprodan (commented):
Please add a test for a GitRepository and its SSH secret in a different namespace

stefanprodan (approved):
LGTM

Thanks @aryan9600

PR #300, opened 2022-01-24
Retire squaremo as a maintainer

squaremo:
This moves me into the "Retired maintainers" category. The format is intended to work with the script https://github.com/fluxcd/community/blob/main/project/get_project_data.

### DIFF
--- a/MAINTAINERS
+++ b/MAINTAINERS
@@ -2,11 +2,11 @@ The maintainers are generally available in Slack at
 https://cloud-native.slack.com in #flux (https://cloud-native.slack.com/messages/CLAJ40HV3)
 (obtain an invitation at https://slack.cncf.io/).
 
-In additional to those listed below, this project shares maintainers
-from the main Flux v2 git repository, as listed in
+This project shares maintainers from the main Flux v2 git repository,
+as listed in
 
     https://github.com/fluxcd/flux2/blob/main/MAINTAINERS
 
-In alphabetical order:
+Retired maintainers:
 
-Michael Bridgen, Weaveworks <michael@weave.works> (github: @squaremo, slack: Michael Bridgen)
+ - Michael Bridgen


MAINTAINERS
```diff
@@ -2,11 +2,15 @@ The maintainers are generally available in Slack at
...
+In additional to those listed as current maintainers below, this project shares maintainers
 from the main Flux v2 git repository, as listed in
 
     https://github.com/fluxcd/flux2/blob/main/MAINTAINERS
 
-In alphabetical order:
+Project maintainers, in alphabetical order:
 
-Michael Bridgen, Weaveworks <michael@weave.works> (github: @squaremo, slack: Michael Bridgen)
+None.
```
kingdonb, line 12: This looks like an issue in need of adjustment, :)

Is there anything in governance about this? (I suppose we'll need someone to volunteer, else there will be nobody for reviewing and merging PRs.)

stefanprodan: This is confusing indeed, it's not like this project is now abandoned and without maintainers, but having `None` here suggests that.

stefanprodan: @squaremo can you please remove the "Project maintainers, in alphabetical order:" section?

squaremo: The alternative is that the initial paragraph refers to a list that does not follow.

squaremo: If you don't read the initial paragraph, either "None" or a missing list will lead you to think there are no maintainers. If you do read it, you will understand that "None" refers to additional maintainers.

stefanprodan: In other repos we don't say `None` we just omit that paragraph: https://github.com/fluxcd/kustomize-controller/blob/main/MAINTAINERS

squaremo: Yep, that's a better formulation: no need to refer to additional maintainers, then not have any.

hiddeco (approved):
:cry: stamp!

(It needs a :recycle: :fish: due to branch protection rule changes)

Issue #309, opened 2022-02-02
Build fails on Apple Silicon


stefanprodan:
After the v0.20 release, running make build/test/vet/etc fails on Apple Silicon with Go 1.17 ARM64. This is due to the libgit2 static build changes introduced in #303

PR #311, opened 2022-02-08
Upgrade libgit2 and fix static builds

aryan9600:
* Bump to golang-with-libgit2:1.1.1.6 to speed up build time when cross compiling. Previous version was compiling in emulation mode instead, which added +10x overhead.
* Ensure that make test is executed against the exact same libraries that will be shipped on the built image.
* Simplify Makefile to reduce its complexity.
* Libgit2 behaviour:
    linux-amd64 download static libraries from the official container image.
    linux-arm64 on top of the above, requires static musl tool chain (automatically downloaded).
    darwin-amd64 and darwin-arm64 download universal static libraries for darwin from https://github.com/fluxcd/golang-with-libgit2 releases.

Fixes: #309 

Co-authored-by: Paulo Gomes <paulo.gomes@weave.works>

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -30,3 +30,42 @@ jobs:
       run: make test
     - name: Verify
       run: make verify
+
+  kind-linux-arm64:
+    # Hosted on Equinix
+    # Docs: https://github.com/fluxcd/flux2/tree/main/.github/runners
+    runs-on: [self-hosted, Linux, ARM64, equinix]
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+      - name: Setup Go
+        uses: actions/setup-go@v2
+        with:
+          go-version: 1.17.x
+      - name: Run tests
+        run: make test
+      - name: Verify
+        run: make verify
+
+  # Runs 'make test' on macos-10.15 to assure development environment for
+  # contributors using MacOS.
+  darwin-amd64:
+    runs-on: macos-10.15
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+      - name: Setup Go
+        uses: actions/setup-go@v2
+        with:
+          go-version: 1.17.x
+      - name: Restore Go cache
+        uses: actions/cache@v1
+        with:
+          path: /home/runner/work/_temp/_github_home/go/pkg/mod
+          key: ${{ runner.os }}-go-${{ hashFiles('**/go.sum') }}
+          restore-keys: |
+            ${{ runner.os }}-go-
+      - name: Run tests
+        run: make test
+      - name: Verify
+        run: make verify
--- a/Dockerfile
+++ b/Dockerfile
@@ -3,56 +3,96 @@ ARG GO_VERSION=1.17
 ARG XX_VERSION=1.1.0
 
 ARG LIBGIT2_IMG=ghcr.io/fluxcd/golang-with-libgit2
-ARG LIBGIT2_TAG=libgit2-1.1.1-4
+ARG LIBGIT2_TAG=libgit2-1.1.1-6
 
-FROM --platform=linux/amd64 ${LIBGIT2_IMG}:${LIBGIT2_TAG} as build-amd64
-FROM --platform=linux/arm64 ${LIBGIT2_IMG}:${LIBGIT2_TAG} as build-arm64
-FROM --platform=linux/arm/v7 ${LIBGIT2_IMG}:${LIBGIT2_TAG} as build-armv7
+FROM ${LIBGIT2_IMG}:${LIBGIT2_TAG} AS libgit2-libs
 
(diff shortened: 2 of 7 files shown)


.github/workflows/build.yaml
```diff
@@ -30,3 +30,48 @@ jobs:
...
+      - name: Setup Go
+        uses: actions/setup-go@v2
+        with:
+          go-version: 1.17.x
+      - name: Run tests
+        run: |
+          mkdir tmp-download; cd tmp-download; go mod init go-download;
+          GOBIN="${GITHUB_WORKSPACE}/build/gobin" go install sigs.k8s.io/controller-runtime/tools/setup-envtest@latest
+          GOBIN="${GITHUB_WORKSPACE}/build/gobin" go install sigs.k8s.io/controller-tools/cmd/controller-gen@v0.7.0
+          GOBIN="${GITHUB_WORKSPACE}/build/gobin" go install github.com/ahmetb/gen-crd-api-reference-docs@v0.3.0
```
pjbgf, line 50: When building against the arm64 runners for SC, only the `setup-envtest` did not work via Makefile. Did you also have issues with `controller-gen` and `gen-crd-api-reference-docs`?

Makefile
```diff
@@ -148,7 +192,7 @@ docker-build:  ## Build the Docker image
 	docker buildx build \
 		--build-arg LIBGIT2_IMG=$(LIBGIT2_IMG) \
 		--build-arg LIBGIT2_TAG=$(LIBGIT2_TAG) \
-		--platform=$(BUILD_PLATFORMS) \
+		--platform=linux/amd64 \
```
pjbgf, line 195: We probably need to revert this:

```suggestion
		--platform=$(BUILD_PLATFORMS) \
```

hack/install-libraries.sh
```diff
@@ -0,0 +1,98 @@
...
+    if [ -d "./build/libgit2/${TAG}" ]; then
+        echo "Skipping libgit2 setup as it already exists"
+        exit 0
+    fi
+
+    mkdir -p "./build/libgit2"
+    if [[ $OSTYPE == 'darwin'* ]]; then
+        # For MacOS development environments, download the amd64 static libraries released from from golang-with-libgit2.
+
+        #TODO: update URL with official URL + TAG:
```
pjbgf, line 49:
```suggestion
```

pjbgf (changes requested):
Nice one, @aryan9600. Just a few nits from me, apart from that really good stuff. :+1:

.github/workflows/build.yaml
aryan9600: yes: https://github.com/fluxcd/image-automation-controller/runs/5110912422?check_suite_focus=true

stefanprodan: Ca we set `GOBIN` for Make and get rid of the duplicate install? The issue with this is that the controller-gen version will diverge from the one in the Makefile.

Dockerfile
```diff
@@ -2,58 +2,98 @@ ARG BASE_VARIANT=alpine
 ARG GO_VERSION=1.17
 ARG XX_VERSION=1.1.0
 
-ARG LIBGIT2_IMG=ghcr.io/fluxcd/golang-with-libgit2
-ARG LIBGIT2_TAG=libgit2-1.1.1-4
```
pjbgf, line 6: This is required for release purposes xref: https://github.com/fluxcd/source-controller/pull/574

Makefile
```diff
@@ -159,43 +203,48 @@ docker-deploy:	## Set the Docker image in-cluster
...
 gen-crd-api-reference-docs:
 	$(call go-install-tool,$(GEN_CRD_API_REFERENCE_DOCS),github.com/ahmetb/gen-crd-api-reference-docs@v0.3.0)
 
-ENVTEST_ASSETS_DIR=$(shell pwd)/testbin
+ENVTEST_ASSETS_DIR=$(BUILD_DIR)/testbin
 ENVTEST_KUBERNETES_VERSION?=latest
 install-envtest: setup-envtest
 	mkdir -p ${ENVTEST_ASSETS_DIR}
 	$(ENVTEST) use $(ENVTEST_KUBERNETES_VERSION) --arch=$(ENVTEST_ARCH) --bin-dir=$(ENVTEST_ASSETS_DIR)
+	chmod -R u+w $(BUILD_DIR)
```
pjbgf, line 232: This may conflict with fuzzing, which creates files that are owned by root. I would rather just reset chmod of things that are set to 0555 by setenvtest:

```suggestion
	chmod -R u+w $(BUILD_DIR)/testbin
```

pjbgf (changes requested):
It is looking really good, thank you @aryan9600. The only change we must do before merging is to reset the img/tag on the Dockerfile, otherwise this will cause the release workflow to fail.

pjbgf (approved):
Great stuff @aryan9600, thank you! LGTM

hiddeco (approved):

Issue #320, opened 2022-02-21
Update libgit2 to 1.3.0


pjbgf:
Benefits and impacted issues:
- Improved error messages for git with meaningful information (opposed to “User error” or “Certificate error”)
  - https://github.com/fluxcd/image-automation-controller/issues/298
- Support for ed25519 as SSH hostkey verification
  - https://github.com/fluxcd/image-automation-controller/issues/298
- Support for IPv6.

Depends on:
- https://github.com/fluxcd/source-controller/pull/573
- https://github.com/fluxcd/source-controller/issues/587

PR #321, opened 2022-02-22
Update libgit2 to 1.3.0

pjbgf:
Update source controller to `v0.21.3` (https://github.com/fluxcd/source-controller/pull/589) and update libgit2 to `v1.3.0`.

Fixes https://github.com/fluxcd/image-automation-controller/issues/320


controllers/update_test.go
```diff
@@ -1413,3 +1435,77 @@ func mergeBranchIntoHead(repo *git.Repository, pushBranch string) {
 	})
 	Expect(err).NotTo(HaveOccurred())
 }
+
+// copied from source-controller/pkg/git/libgit2/checkout.go
+func commitFile(repo *git2go.Repository, path, content string, time time.Time) (*git2go.Oid, error) {
```
stefanprodan, line 1440: I think we should rewrite the commit logic using libgit2 https://github.com/fluxcd/image-automation-controller/blob/544e4870393435ec07c078c69dc9749149bc60c4/controllers/imageupdateautomation_controller.go#L571

pjbgf: I rewrote the gitInitRepo with libgit2, which was enough to remove the need for the adhoc fix.

Can we go ahead with the libgit2 to `1.3.0` on this PR and then any further rewrite/replacement of go-git we can deal on a separate PR.

stefanprodan (approved):
LGTM

Thanks @pjbgf

Issue #282, opened 2021-12-17
Controller stops reconciling, needs restart


stefanprodan:
Reported here: https://github.com/fluxcd/flux2/discussions/2219

Having an automation that should reconcile every 7 minutes:

```
apiVersion: image.toolkit.fluxcd.io/v1beta1
kind: ImageUpdateAutomation
metadata:
  name: flux-system
  namespace: flux-system
spec:
  git:
    checkout:
      ref:
        branch: master
    commit:
      author:
        email: me@example.com
        name: me
      messageTemplate: '{{range .Updated.Images}}{{println .}}{{end}}'
    push:
      branch: master
  interval: 7m0s
  sourceRef:
    kind: GitRepository
    name: flux-system
  update:
    path: ./staging
    strategy: Setters
```

The reconciliation stoped two days ago for unknown reasons:
```console
$ date
Fri Dec 17 16:25:48 EET 2021

$ flux get image update

NAME       	READY	MESSAGE                                                     	LAST RUN                 	SUSPENDED
flux-system	True 	no updates made; last commit 8574614 at 2021-12-14T22:47:08Z	2021-12-15T08:15:01-07:00	False
```

jwerre:
This coincided with outages at AWS in us-west-2 where I'm running Kubernetes on EKS. My entire cluster went down on that day for about 10 minutes and when it came back up I ran into this issue.

bondido:
Hi,
While waiting for permanent fix, are there any advice how to detect "stucked" image-automation-controller case?
Are there any metrics or logs we should monitor?
We have been suffering the issue (~couple of times per week) for some time and the only alert is from users who pushed their images to container registry and didn't notice cluster deployment for quite some time.

Another observation is that in most cases restarting image-automation-controller is sufficient, but there where two times when we needed to also restart source-controller.

jwerre:
@bondido how about something like this:

```bash
kubectl get pod -n flux-system

NAME                                          READY   STATUS    RESTARTS   AGE
helm-controller-96dc99bfd-p9g4q               1/1     Running   0          9d
image-automation-controller-64c955c59-ckbft   1/1     Running   0          9d
image-reflector-controller-55fb7f765d-cr8zn   1/1     Running   0          9d
kustomize-controller-7bc878f768-xf2xb         1/1     Running   0          9d
notification-controller-945795558-q8nht       1/1     Running   0          9d
source-controller-65665fd68f-n7qqz            1/1     Running   0          9d
```

[...]

Issue #209, opened 2021-08-03
image-automation-controller not reconnecting after operation timed out


minh-nguyenquang:
### Describe the bug

image-automation-controller doesn't reconnect to github after operation timed out. I have to delete the pod to restart.
Below is the log from image-automation-controller.
```
{"level":"error","ts":"2021-07-22T14:57:58.859Z","logger":"controller-runtime.manager.controller.imageupdateautomation","msg":"Reconciler error","reconciler group":"image.toolkit.fluxcd.io","reconciler kind":"ImageUpdateAutomation","name":"flux-system","namespace":"flux-system","error":"unable to clone 'https://github.com/CoverGo/k8s-fleet', error: failed to connect to github.com: Operation timed out"}
{"level":"error","ts":"2021-07-22T15:03:04.528Z","logger":"controller-runtime.manager.controller.imageupdateautomation","msg":"Reconciler error","reconciler group":"image.toolkit.fluxcd.io","reconciler kind":"ImageUpdateAutomation","name":"flux-system","namespace":"flux-system","error":"unable to clone 'https://github.com/CoverGo/k8s-fleet', error: failed to connect to github.com: Operation timed out"}
```

### Steps to reproduce

I don't know how to reproduce because operation timed out can happen anytime

### Expected behavior

image-automation-controller can reconnect automatically.

### OS / Distro

N/A

### Flux version

0.16.1

### Flux check

► checking prerequisites
✔ kubectl 1.21.0 >=1.18.0-0
✔ Kubernetes 1.18.8-aliyun.1 >=1.16.0-0
► checking controllers
✔ helm-controller: deployment ready
► ghcr.io/fluxcd/helm-controller:v0.11.1
✔ image-automation-controller: deployment ready
► ghcr.io/fluxcd/image-automation-controller:v0.14.0
✔ image-reflector-controller: deployment ready
► ghcr.io/fluxcd/image-reflector-controller:v0.11.0
✔ kustomize-controller: deployment ready
► ghcr.io/fluxcd/kustomize-controller:v0.13.2
✔ notification-controller: deployment ready
► ghcr.io/fluxcd/notification-controller:v0.15.0
✔ source-controller: deployment ready
► ghcr.io/fluxcd/source-controller:v0.15.3
✔ all checks passed

### Git provider

github

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

stefanprodan:
[...]

PR #326, opened 2022-03-18
Add experimental managed transport for libgit2 operations

pjbgf:
`libgit2` network operations are blocking and do not provide timeout nor context capabilities, leading to users reporting that image-automation-controller hanging indefinitely.

By using managed transport, golang primitives such as http.Transport and net.Dial can be used to ensure timeouts are enforced.

This will initially be guarded behind a feature switch via environment variable `EXPERIMENTAL_GIT_TRANSPORT`.

Extends from https://github.com/fluxcd/source-controller/pull/606
Affected issues: https://github.com/fluxcd/image-automation-controller/issues/209 https://github.com/fluxcd/image-automation-controller/issues/286 https://github.com/fluxcd/image-automation-controller/issues/282

### DIFF
--- a/controllers/imageupdateautomation_controller.go
+++ b/controllers/imageupdateautomation_controller.go
@@ -544,7 +544,8 @@ func switchBranch(repo *libgit2.Repository, pushBranch string) error {
 		}
 		defer head.Free()
 
-		_, err = repo.CreateBranch(pushBranch, head, false)
+		branch, err := repo.CreateBranch(pushBranch, head, false)
+		defer branch.Free()
 		return err
 	}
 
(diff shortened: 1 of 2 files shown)


controllers/imageupdateautomation_controller.go
```diff
@@ -247,6 +248,35 @@ func (r *ImageUpdateAutomationReconciler) Reconcile(ctx context.Context, req ctr
 		return failWithError(err)
 	}
 
+	repositoryURL := origin.Spec.URL
+	// managed GIT transport only affects the libgit2 implementation
```
hiddeco, line 252: Think that with recent changes, the implementation always _is_ libgit2?

stefanprodan (approved):
LGTM

Thanks @pjbgf 🏅

PR #351, opened 2022-04-13
Add flag to allow configuration of ssh kex algos

aryan9600:
Adds a runtime flag to specify which key exchange algorithms to use for ssh.

Ref: https://github.com/fluxcd/flux2/issues/2610


pjbgf (approved):
@aryan9600 great addition!

LGTM

PR #362, opened 2022-05-12
git: Enable managed transport by default

pjbgf:
`GitManagedTransport` is a new feature gate that enables the use of Managed Transport. 
This replaces the previous `EXPERIMENTAL_GIT_TRANSPORT` environment variable that was
used for the same result. 

This is now an `opt-out` feature, which can be disabled by starting the
controller with the argument '--feature-gates=GitManagedTransport=false'.

---
This is part of a series of changes that happened on source-controller that needs to be synced with IAC.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -13,7 +13,7 @@ require (
 	github.com/fluxcd/pkg/apis/acl v0.0.3
 	github.com/fluxcd/pkg/apis/meta v0.13.0
 	github.com/fluxcd/pkg/gittestserver v0.5.2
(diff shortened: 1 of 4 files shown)


pjbgf:
> Do we still make use of the env var? I don't see it removed in this PR.

@stefanprodan  With the new way of initialising this would be enough, however, the version of SC this uses still lingers on that environment variable to check the Managed Transport status. We will need to first get it merged and released on SC before we can do the same here.
I created the PR for SC and will keep this as draft until we are ready to update/merged it.

pjbgf:
superseded by #369

PR #368, opened 2022-05-26
Update dependencies

pjbgf:
Update dependencies:
- fluxcd/pkg/apis/meta v0.14.1
- fluxcd/pkg/runtime v0.16.1
- github.com/ProtonMail/go-crypto 0.0.0-20220517143526-88bb52951d5b
- github.com/fluxcd/pkg/gittestserver 0.5.4
- github.com/fluxcd/pkg/ssh 0.4.0
- github.com/google/go-containerregistry 0.9.0
- golang.org/x/crypto 0.0.0-20220525230936-793ad666bf5e
- sigs.k8s.io/kustomize/kyaml 0.13.7
- gopkg.in/yaml.v3 v3.0.0

And the GH actions.

Depends on #369

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -13,13 +13,13 @@ jobs:
     runs-on: ubuntu-latest
     steps:
     - name: checkout
-      uses: actions/checkout@v2
+      uses: actions/checkout@v3
     - name: Setup Go
-      uses: actions/setup-go@v2
+      uses: actions/setup-go@v3
       with:
         go-version: 1.17.x
     - name: Restore go cache
-      uses: actions/cache@v1
+      uses: actions/cache@v3
       with:
         # the ff is mounted into the container as ~/go/pkg/mod
         path: /home/runner/work/_temp/_github_home/go/pkg/mod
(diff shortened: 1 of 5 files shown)


go.mod
```diff
@@ -6,33 +6,38 @@ replace github.com/fluxcd/image-automation-controller/api => ./api
...
+	k8s.io/api v0.24.0
+	k8s.io/apimachinery v0.24.0
+	k8s.io/client-go v0.24.0
+	k8s.io/kube-openapi v0.0.0-20220401212409-b28bf2818661
 	sigs.k8s.io/controller-runtime v0.11.2
-	sigs.k8s.io/kustomize/kyaml v0.13.6
+	sigs.k8s.io/kustomize/kyaml v0.13.7
 )
 
+require github.com/sosedoff/gitkit v0.3.0 // indirect
```
stefanprodan, line 34: Let's use fluxcd/gitkit

stefanprodan:
I think we should wait until we release SC so we can get rid of the gitkit replacement, we need to switch SC to `fluxcd/gitkit` first.

stefanprodan (approved):

Issue #370, opened 2022-05-27
Test Managed Transport


pjbgf:
The Git related tests must be refactored to ensure the Managed Transport is being used.
Given that a lot of the testing logic is based on `git2go`, by enabling the Manage Transport of the controller we are also changing the behaviour of the tests themselves.

Extra points to consider:
- `libgit2` checkout func now have the side effect of removing the Transport Options even if they were added by the caller. We probably want to change this. 
- Run both managed and unmanaged tests whilst we support the latter.

pjbgf:
Managed transport has been deprecated since [v0.24.0](https://github.com/fluxcd/image-automation-controller/blob/main/CHANGELOG.md#0240) and therefore this is no longer required.

PR #330, opened 2022-03-23
Fix bug when pushing into different branch

pjbgf:
Changes replacing go-git with git2go introduced a bug in which pushes into new branches squashes all commits into one.

Relates to: https://github.com/fluxcd/image-automation-controller/issues/329

PR #373, opened 2022-06-02
git: refactor tests to use managed transports and fix switchToBranch

aryan9600:
This PR is a follow up to #369.
It refactors the tests to use managed transports, since that's the default now. 
It also fixes a regression introduced in #330, which would result in the controller failing to push commits, if the remote is ahead of the local branch. This was partially addressed in #369, by fetching the commits, but it failed to update the local branch target to be at the latest commit.

Fixes most of #370

### DIFF
--- a/controllers/git_test.go
+++ b/controllers/git_test.go
@@ -10,6 +10,9 @@ import (
 
 	"github.com/go-logr/logr"
 	libgit2 "github.com/libgit2/git2go/v33"
+	"k8s.io/apimachinery/pkg/types"
+
+	. "github.com/onsi/gomega"
 
 	"github.com/fluxcd/pkg/gittestserver"
 	"github.com/fluxcd/source-controller/pkg/git/libgit2/managed"
@@ -129,12 +132,12 @@ func TestPushRejected(t *testing.T) {
 	}
 
 	// this is currently defined in update_test.go, but handy right here ..
-	if err = initGitRepo(gitServer, "testdata/appconfig", "master", "/appconfig.git"); err != nil {
+	if err = initGitRepo(gitServer, "testdata/appconfig", "test", "/appconfig.git"); err != nil {
 		t.Fatal(err)
 	}
 
 	repoURL := gitServer.HTTPAddressWithCredentials() + "/appconfig.git"
-	repo, err := clone(repoURL, "master")
+	repo, err := clone(repoURL, "test")
 	if err != nil {
 		t.Fatal(err)
 	}
@@ -148,7 +151,7 @@ func TestPushRejected(t *testing.T) {
 	repo.Remotes.SetUrl("origin", transportOptsURL)
 
 	// This is here to guard against push in general being broken
-	err = push(context.TODO(), repo.Workdir(), "master", repoAccess{})
+	err = push(context.TODO(), repo.Workdir(), "test", repoAccess{})
 	if err != nil {
 		t.Fatal(err)
 	}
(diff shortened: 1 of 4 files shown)


controllers/imageupdateautomation_controller.go
```diff
@@ -784,15 +784,28 @@ func switchToBranch(repo *libgit2.Repository, ctx context.Context, branch string
...
-	}
-	if localBranch == nil {
-		return fmt.Errorf("cannot create local branch '%s': %w", branch, err)
+		// We could've done something like:
+		// lb, err := repo.CreateBranch()
+		// if lb != nil {
+		//	 defer lb.Free()
+		//	 localBranch = lb.Reference
+		// }
+		// But for some reasone, calling `lb.Free()` under any circumstances, causes a really
```
hiddeco, line 798:
```suggestion
		// But for some reason, calling `lb.Free()` under any circumstances, causes a really
```

controllers/update_test.go
```diff
@@ -1123,40 +1157,53 @@ func mockSignature(time time.Time) *git2go.Signature {
...
-func clone(repoURL, remoteName, branchName string) (*git2go.Repository, error) {
+func clone(repoURL, branchName string) (*git2go.Repository, error) {
 	dir, err := os.MkdirTemp("", "iac-clone-*")
 	if err != nil {
 		return nil, err
 	}
+	transportOptsURL := "http://" + randStringRunes(5)
+	managed.AddTransportOptions(transportOptsURL, managed.TransportOptions{
+		TargetURL: repoURL,
+	})
```
darkowlzz, line 1168: Since transport options now accept context, `clone()` can accept a context and pass it to transport option.

controllers/update_test.go
```diff
@@ -1123,40 +1157,53 @@ func mockSignature(time time.Time) *git2go.Signature {
...
 		},
-		FetchOptions: git2go.FetchOptions{
-			RemoteCallbacks: git2go.RemoteCallbacks{
-				CertificateCheckCallback: func(cert *git2go.Certificate, valid bool, hostname string) error {
-					return nil
-				},
-			},
-		},
 	}
+	repo, err := git2go.Clone(transportOptsURL, dir, opts)
```
darkowlzz, line 1178: This can have an error check.

controllers/update_test.go
```diff
@@ -480,8 +480,12 @@ func TestImageAutomationReconciler_e2e(t *testing.T) {
 
 		t.Run("PushSpec", func(t *testing.T) {
 			// Clone the repo locally.
-			localRepo, err := clone(cloneLocalRepoURL, "origin", branch)
+			localRepo, err := clone(cloneLocalRepoURL, branch)
 			g.Expect(err).ToNot(HaveOccurred(), "failed to clone git repo")
+			defer localRepo.Free()
+			origin, err := localRepo.Remotes.Lookup("origin")
+			g.Expect(err).ToNot(HaveOccurred(), "failed to clone git repo")
+			defer origin.Free()
```
darkowlzz, line 488: If this origin lookup needs to be performed after a clone to make sure clone happened successfully, maybe it'd be better to move this check within the clone function.

controllers/update_test.go
```diff
@@ -872,10 +875,39 @@ func compareRepoWithExpected(g *WithT, repoURL, branch, fixture string, changeFi
...
+// back to the actual url. Callers are expected to call this function in a deferred manner.
+func configureManagedTransportOptions(repo *libgit2.Repository, repoURL string) (func(), error) {
+	if repoURL == "" {
+		origin, err := repo.Remotes.Lookup(originRemote)
+		if err != nil {
+			return nil, err
+		}
+		defer origin.Free()
+		repoURL = origin.Url()
+	}
```
darkowlzz, line 891: Based on the usage of this function, it looks like it'd be better to not accept `repoURL` and derive it out of the `repo` always.

darkowlzz: And with that, this function can be renamed to `configureManagedTransportOptionsForRepo` or something short and better if you can think of.

controllers/update_test.go
```diff
@@ -872,10 +875,39 @@ func compareRepoWithExpected(g *WithT, repoURL, branch, fixture string, changeFi
...
+	transportOptsURL := "http://" + randStringRunes(5)
```
darkowlzz, line 892: Maybe the scheme should be based on the URL?

controllers/update_test.go
```diff
@@ -1123,40 +1157,53 @@ func mockSignature(time time.Time) *git2go.Signature {
...
 
+	cleanup, err := configureManagedTransportOptions(repo, "")
+	if err != nil {
+		panic(err)
+	}
+	defer cleanup()
+
+	origin, err := repo.Remotes.Lookup("origin")
+	if err != nil {
+		panic("origin not set")
```
darkowlzz, line 1201: Don't think we need to introduce any panic. Just `g.Expect()` assertions should be enough to stop further execution.

controllers/update_test.go
```diff
@@ -1385,11 +1438,17 @@ func testWithCustomRepoAndImagePolicy(
 
 	// Clone the repo.
 	repoURL := gitServer.HTTPAddressWithCredentials() + repositoryPath
-	localRepo, err := clone(repoURL, "origin", args.branch)
+	localRepo, err := clone(repoURL, args.branch)
 	g.Expect(err).ToNot(HaveOccurred(), "failed to clone git repo")
+	defer localRepo.Free()
+
+	origin, err := localRepo.Remotes.Lookup("origin")
```
darkowlzz, line 1445: Can just use `originRemote` here and some of the other code above and below as in various other place.

controllers/git_test.go
```diff
@@ -128,21 +132,26 @@ func TestPushRejected(t *testing.T) {
...
-	// This is here to guard against push in general being broken
-	err = push(context.TODO(), repo.Workdir(), "main", repoAccess{
-		url:  repoURL,
-		auth: nil,
+	transportOptsURL := "http://" + randStringRunes(5)
+	managed.AddTransportOptions(transportOptsURL, managed.TransportOptions{
+		TargetURL: repoURL,
 	})
+	defer managed.RemoveTransportOptions(transportOptsURL)
+	repo.Remotes.SetUrl("origin", transportOptsURL)
```
darkowlzz, line 151: Looks like we can use the newly introduced `configureManagedTransportOptions()` to do the above.

controllers/git_test.go
```diff
@@ -128,21 +132,26 @@ func TestPushRejected(t *testing.T) {
 	}
 
 	// this is currently defined in update_test.go, but handy right here ..
-	if err = initGitRepo(gitServer, "testdata/appconfig", "main", "/appconfig.git"); err != nil {
+	if err = initGitRepo(gitServer, "testdata/appconfig", "test", "/appconfig.git"); err != nil {
```
darkowlzz, line 135: Would be nice to capture the reason for using a different branch in commit description.

controllers/git_test.go
```diff
@@ -154,11 +163,98 @@ func TestPushRejected(t *testing.T) {
...
+	defer head.Free()
+	target := head.Target()
+
+	// register transport options and update remote to transport url
+	transportOptsURL := "http://" + randStringRunes(5)
+	managed.AddTransportOptions(transportOptsURL, managed.TransportOptions{
+		TargetURL: repoURL,
+	})
+	defer managed.RemoveTransportOptions(transportOptsURL)
+	repo.Remotes.SetUrl("origin", transportOptsURL)
```
darkowlzz, line 198: Can use `configureManagedTransportOptions()` here as well.

controllers/update_test.go
darkowlzz: It may be the case that the origin URL lacks http credentials, that should be verified. But if that's the case, the `repoURL` argument can be kept and renamed appropriately, maybe `urlWithCreds` or something better.

controllers/imageupdateautomation_controller.go
```diff
@@ -784,15 +784,28 @@ func switchToBranch(repo *libgit2.Repository, ctx context.Context, branch string
...
+		// nasty crash. Since not calling `lb.Free()` could result in memory leaks, we avoid
+		// assigning a variable like `lb` and get the branch reference manually.
+		_, err := repo.CreateBranch(branch, commit, false)
```
darkowlzz, line 801: I don't think if we just skip assigning the branch to a variable, there won't be a memory leak. Since C doesn't have GC and it's Go's responsibility to free it. Not taking the responsibility of freeing the memory would not free the memory automatically.

darkowlzz: We do have similar code in SC https://github.com/fluxcd/source-controller/blob/dc1037d9621b80f1639f6ed9c4bfe98db86d24b9/pkg/git/libgit2/checkout.go#L169.

controllers/imageupdateautomation_controller.go
```diff
@@ -810,6 +823,11 @@ func switchToBranch(repo *libgit2.Repository, ctx context.Context, branch string
 		return fmt.Errorf("cannot checkout tree for branch '%s': %w", branch, err)
 	}
 
+	_, err = localBranch.SetTarget(commit.Id(), "")
```
darkowlzz, line 826: Similar to above, the reference returned here has to be freed.

controllers/imageupdateautomation_controller.go
darkowlzz: So, I managed to reproduce this running the test.
Using the reference seems to cause the crash. Let's not use the reference but at least assign the returned branch and free it.

controllers/git_test.go
aryan9600: adding a comment instead

darkowlzz (approved):
LGTM!
Thanks a lot for the test improvements.

hiddeco (approved):

PR #384, opened 2022-06-08
Update github.com/fluxcd/source-controller v0.25.5

pjbgf:
This version of source-controller fixes an issue when checking `known_hosts` by which previous attempted hashes would overflow into the subsequent hash input, rendering the results invalid. This would only impact cases in which multiple hashes were provided for a single target host.

Fixes https://github.com/fluxcd/image-automation-controller/issues/378.


stefanprodan (changes requested):
Please link to https://github.com/fluxcd/image-automation-controller/issues/378 and explain in the description how this fixed a regression bug.

stefanprodan (approved):
LGTM

Thanks @pjbgf

PR #388, opened 2022-06-12
Add new flag --ssh-hostkey-algos

pjbgf:
The new flag allow users to set the list of hostkey algorithms to use for ssh connections, enabling them to ensure specific are/aren't used.

Aligns with source controller flags.


aryan9600 (approved):
LGTM!

PR #392, opened 2022-06-14
Add logging when adding files to index

aryan9600:


pjbgf:
PS: This PR is for debug purposes only and is to be merged into the `change-index` branch.

Issue #405, opened 2022-07-07
Reconciler error: early EOF


Sonlis:
I did setup an image repository, image policy and image update automation. The image repository and image policy work correctly, and find the latest image based on the policy. 
However, the image-automation-controller returns this error: 
```
{"level":"error","ts":"2022-07-07T13:16:53.005Z","logger":"controller.imageupdateautomation","msg":"Reconciler error","reconciler group":"image.toolkit.fluxcd.io","reconciler kind":"ImageUpdateAutomation","name":"license-service","namespace":"sharpershape","error":"early EOF"}
```
The ImageUpdateAutomation sourceRef is as follows: 
```
apiVersion: image.toolkit.fluxcd.io/v1beta1
kind: ImageUpdateAutomation
metadata:
  name: license-service
spec:
  interval: 1m0s
  sourceRef:
    kind: GitRepository
    name: flux-dev-source-repository
    namespace: admin
  git:
    checkout:
      ref:
        branch: master
    commit:
      author:
        email: fluxcdbot@users.noreply.github.com
        name: fluxcdbot
      messageTemplate: '{{range .Updated.Images}}{{println .}}{{end}}'
    push:
      branch: master
---
apiVersion: source.toolkit.fluxcd.io/v1beta2
kind: GitRepository
metadata:
  name: flux-dev-source-repository
  namespace: admin
spec:
  interval: 1m0s
  ref:
    branch: master
  secretRef:
    name: flux-system
  url: ssh://git@github.com/companyName/flux-dev-v2
``` 

This gitRepository source is also used by the source-controller and can fetch properly the repository. I have also checked that the ssh-key used by this gitRepository has both read and write rights, which it does (we however use ssh-keys for the user flux, not one per repository. It has then access to multiple repository with only one ssh key).
The error is a bit obscure, and the only similar I could find in the github issues was solved by having write permissions on the ssh key, which is already setup.

image-automation-controller version: v0.23.4

Sonlis:
We don't bootstrap through the CLI, we use terraform and generate our own SSH keys and input them into a secret.
The thing is, the ssh key has read/write permissions and we still get this error:
<img width="767" alt="image" src="https://user-images.githubusercontent.com/48835068/177932289-351b34b3-b9c6-4b13-a009-b265fa74cfc6.png">

Sonlis:
Turns out I forgot to add the flux user to the repository, so it couldn't manage it. Sorry for the inconvenience, managed to have it working now.

PR #410, opened 2022-07-20
Enrich `early EOF` error message

pjbgf:
Some Git servers may close the connection abruptly when the user does not have write access to a repository whilst trying to do a push operation. The result for the end users is an `early EOF` message, which is not very helpful.

This PR changes the error message to: `early EOF (the SSH key may not have write access to the repository)`.

Fixes https://github.com/fluxcd/image-automation-controller/issues/405.


controllers/git_test.go
```diff
@@ -172,6 +173,54 @@ func TestPushRejected(t *testing.T) {
...
+	defer repo.Free()
+
+	cleanup, err := configureTransportOptsForRepo(repo, access.auth)
+	g.Expect(err).ToNot(HaveOccurred())
+
+	defer cleanup()
+
+	err = push(context.TODO(), repo.Workdir(), "test", access)
+	g.Expect(err).To(HaveOccurred())
+	g.Expect(err.Error()).To(ContainSubstring("early EOF"))
```
aryan9600, line 221: Since this test is supposed to test the enrichment of the "early EOF" error messages, it'd make more sense to check for the additional message that we explicitly add.

controllers/imageupdateautomation_controller.go
```diff
@@ -887,6 +887,9 @@ func push(ctx context.Context, path, branch string, access repoAccess) error {
 		ProxyOptions:    libgit2.ProxyOptions{Type: libgit2.ProxyTypeAuto},
 	})
 	if err != nil {
+		if err.Error() == "early EOF" {
+			return fmt.Errorf("early EOF (the SSH key may not have write access to the repository)")
+		}
```
aryan9600, line 892: maybe this would be better?
 
```suggestion
		if strings.Contains(err.Error(), "early EOF") {
			return fmt.Errorf("%w (the SSH key may not have write access to the repository)", err)
		}
```

aryan9600 (changes requested):

stefanprodan (approved):
LGTM

Thanks @pjbgf

aryan9600 (approved):
lgtm! 💯

hiddeco (approved):

PR #412, opened 2022-08-09
Decommission libgit2 Unmanaged Transport

pjbgf:
Decomissions libgit2 Unmanaged Transport and updates the dependencies below:
- k8s.io/api to version 0.24.3.
- k8s.io/apimachinery to version 0.24.3.
- k8s.io/client-go to version 0.24.3.
- github.com/fluxcd/image-reflector-controller/api to version 0.20.0.
- github.com/fluxcd/source-controller to version 0.26.0.
- k8s.io/kube-openapi to version v0.0.0-20220803164354-a70c9af30aea.
- github.com/ProtonMail/go-crypto to version v0.0.0-20220730123233-d6ffb7692adf.
- sigs.k8s.io/kustomize/kyaml to version v0.13.9.


stefanprodan (approved):
LGTM 

Thanks @pjbgf

aryan9600 (changes requested):
`ImageUpdateAutomationReconciler` needs to be updated to have `Libgit2TransportInitialized` as a field (ref: https://github.com/fluxcd/source-controller/blob/v0.26.0/controllers/gitrepository_controller.go#L120) and the reconcile logic needs to verify whether libgit2 transports have been initialized before performing any git operations.

The `GitManagedTransport` feature needs to be removed from `internal/features/features.go`

pjbgf:
@aryan9600 just removed `GitManagedTransport`, thanks for pointing it out. 

For `Libgit2TransportInitialized` as currently the controller only supports `libgit2` this PR exits early in `main.go` if it can't initialise `libgit2`. We can review this approach once we all git implementation code is in `fluxcd/pkg`.

pjbgf:
This has been tested against GitHub, GitLab and BitBucket. All worked as expected.

PR #430, opened 2022-09-02
refactor reconciler to use fluxcd/pkg/git

aryan9600:
Refactor reconciler to use `fluxcd/pkg/git` and `fluxcd/pkg/git/libgit2`. Consequently, remove all git related helper methods and tests.


controllers/imageupdateautomation_controller.go
```diff
@@ -265,54 +261,39 @@ func (r *ImageUpdateAutomationReconciler) Reconcile(ctx context.Context, req ctr
...
+	}
+	defer lgc.Close()
+
+	opts := git.CloneOptions{}
+	if ref != nil {
+		opts.Tag = ref.Tag
+		opts.SemVer = ref.SemVer
+		opts.Commit = ref.Commit
+		opts.Branch = ref.Branch
 	}
```
pjbgf, line 281: This could be a helper func, as we will need to convert ref to `CloneOptions` here and in tests.

controllers/suite_test.go
```diff
@@ -65,7 +65,9 @@ func TestMain(m *testing.M) {
 		filepath.Join("testdata", "crds"),
 	))
 
-	managed.InitManagedTransport()
+	if err := transport.InitManagedTransport(); err != nil {
+		panic(fmt.Sprintf("Failed to initialize libgit2 managed transport: %v", err))
```
pjbgf, line 69:
```suggestion
		panic(fmt.Sprintf("failed to initialize libgit2 managed transport: %v", err))
```

aryan9600: I kept it as "Failed" to match the panic statements below

PR #431, opened 2022-09-05
Aligns output with source-controller on no-ops

pjbgf:
No-op reconciliations now yield a message such as below even when log level is set to `info`, which aligns with source-controller behaviour:
```
{"level":"info","ts":"2022-09-05T08:43:59.516Z","msg":"no changes made in working directory; no commit","controller":"imageupdateautomation","controllerGroup":"image.toolkit.fluxcd.io","controllerKind":"ImageUpdateAutomation","imageUpdateAutomation":{"name":"github-branch","namespace":"flux-system"},"namespace":"flux-system","name":"github-branch","reconcileID":"4ee714d8-5986-45fd-9f54-33d882b81cde"}
```

Custom code to record metrics was replaced with `helper.Metrics`.

### DIFF
--- a/controllers/imageupdateautomation_controller.go
+++ b/controllers/imageupdateautomation_controller.go
@@ -35,12 +35,9 @@ import (
 	"github.com/go-logr/logr"
 	libgit2 "github.com/libgit2/git2go/v33"
 	corev1 "k8s.io/api/core/v1"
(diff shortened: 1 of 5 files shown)


stefanprodan (approved):
LGTM

Thanks @pjbgf

PR #446, opened 2022-10-17
use gitrepo implementation for git ops

aryan9600:
Instead of using libgit2 for all Git operations, use the implementation specified in the GitRepository object.

### DIFF
--- a/controllers/imageupdateautomation_controller.go
+++ b/controllers/imageupdateautomation_controller.go
@@ -50,7 +50,8 @@ import (
 	apiacl "github.com/fluxcd/pkg/apis/acl"
 	"github.com/fluxcd/pkg/apis/meta"
 	"github.com/fluxcd/pkg/git"
-	libgit2pkg "github.com/fluxcd/pkg/git/libgit2"
+	"github.com/fluxcd/pkg/git/gogit"
+	"github.com/fluxcd/pkg/git/libgit2"
 	"github.com/fluxcd/pkg/runtime/acl"
 	helper "github.com/fluxcd/pkg/runtime/controller"
 	"github.com/fluxcd/pkg/runtime/events"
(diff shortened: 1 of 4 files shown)


controllers/imageupdateautomation_controller.go
```diff
@@ -255,11 +256,16 @@ func (r *ImageUpdateAutomationReconciler) Reconcile(ctx context.Context, req ctr
 		return failWithError(err)
 	}
 
-	lgc, err := libgit2pkg.NewClient(tmp, authOpts)
+	var gitClient git.RepositoryClient
+	if origin.Spec.GitImplementation == libgit2.ClientName {
```
hiddeco, line 260: Switch statement would seem more appropriate here with a `default` which errors out for an unsupported implementation string?

controllers/imageupdateautomation_controller.go
```diff
@@ -255,11 +256,19 @@ func (r *ImageUpdateAutomationReconciler) Reconcile(ctx context.Context, req ctr
 		return failWithError(err)
 	}
 
-	lgc, err := libgit2pkg.NewClient(tmp, authOpts)
+	var gitClient git.RepositoryClient
+	switch origin.Spec.GitImplementation {
+	case libgit2.ClientName:
+		gitClient, err = libgit2.NewClient(tmp, authOpts)
+	case gogit.ClientName:
```
pjbgf, line 263: The default empty behaviour should still be `go-git`:
```suggestion
	case gogit.ClientName, "":
```

aryan9600: do you mean `libgit2`?

pjbgf: Now we are getting the value from the `spec.gitImplementation`, I think we should ensure that it works in alignment with `source-controller`.

If in any scenario we need to force `libgit2` we can do that through other means.

aryan9600: okay, since an intended side effect of this change is to reduce the number of crashes due to libgit2, i concur 👍

hiddeco: Wouldn't it in general be better to match against the `v1beta2.GoGitImplementation` and `v1beta2.Libgit2Implementation` variables to ensure we match against the declarative API values opposed to the client package?

pjbgf (approved):
LGTM

Thanks @aryan9600! :bow:

PR #452, opened 2022-10-27
gogit: Add new ForceGoGitImplementation FeatureGate

pjbgf:
ForceGoGitImplementation ignores the value set for gitImplementation and ensures that go-git is used for all GitRepository objects. This can be used to confirm that Flux instances won't break if/when the libgit2 implementation was to be deprecated.

When enabled, libgit2 won't be initialized, nor will any git2go cgo code be called.


aryan9600 (changes requested):

aryan9600 (approved):
lgtm!

stefanprodan (approved):

Issue #466, opened 2022-11-29
Multiple empty commits since flux2 v0.37


srgvg:
[Initially discussed in slack](https://cloud-native.slack.com/archives/CLAJ40HV3/p1669739085669969)
```
➞  flux version
flux: v0.37.0
helm-controller: v0.21.0
kustomize-controller: v0.25.0
notification-controller: v0.23.5
source-controller: v0.24.4
```
This started after upgrading to `0.37`:
```
Date:   Fri Nov 25 00:33:30 2022 +0000

    Automated image update flux-system/flux-system

    Images:
    Objects:
    Files:
```
I get such a commit like every 5s

At some point, restarting the image-automation controller stopped that, however some time later it had restarted again. I think I experienced that twice, however not realising exactly enough what was happening, I didn't keep any logs from this controller, unfortunately.

This is the same base git repo that flux kustomize uses, happens on the master branch, hosted in Gitlab. Verified there were no commits in the container image gerenrating git commit that would have warranted any update. The latest actual image version, gitlab pipeline and latest git commit are all related, and several days older than aforementioned flux2 upgrade. All git authentication happens through an SSH keypair.

Relevant automation manifests:
```
apiVersion: v1
items:
- apiVersion: image.toolkit.fluxcd.io/v1beta1
  kind: ImagePolicy
  metadata:
    labels:
      kustomize.toolkit.fluxcd.io/name: mycompany-www-dev
      kustomize.toolkit.fluxcd.io/namespace: flux-system
    name: mycompany-www-dev
    namespace: flux-system
  spec:
    filterTags:
      extract: $datetime
      pattern: ^(?P<datetime>[0-9]{6}.[0-9]{6})-(?P<hash>[a-fA-F0-9]{8})-dev
    imageRepositoryRef:
      name: mycompany-www-dev
    policy:
      numerical:
        order: asc
- apiVersion: image.toolkit.fluxcd.io/v1beta1
  kind: ImagePolicy
  metadata:
    labels:
      kustomize.toolkit.fluxcd.io/name: mycompany-www-prod
      kustomize.toolkit.fluxcd.io/namespace: flux-system
    name: mycompany-www-prod
    namespace: flux-system
  spec:
    filterTags:
      extract: $datetime
      pattern: ^(?P<datetime>[0-9]{6}.[0-9]{6})-(?P<hash>[a-fA-F0-9]{8})-prod
    imageRepositoryRef:
      name: mycompany-www-prod
    policy:
[...]

PR #470, opened 2022-12-20
Block the creation of empty commits

pjbgf:
In specific cases, the update mechanism may lead to no files being changed. The changes short-circuit the process to exit early when 0 files have been modified.

go-git has been changed to return an error when an empty commit is detected, which is used as a last resort.

When the push branch is different than the clone branch, the controller will now download all branch references from the Git server, in order to enable a single fetch operation.

Fixes https://github.com/fluxcd/image-automation-controller/issues/466.


controllers/imageupdateautomation_controller.go
```diff
@@ -358,21 +366,28 @@ func (r *ImageUpdateAutomationReconciler) Reconcile(ctx context.Context, req ctr
...
-			Message: message,
-		},
-		repository.WithSigner(signingEntity),
-	); err != nil {
-		if err != git.ErrNoStagedFiles {
+			repository.WithSigner(signingEntity),
+		)
+	} else {
+		err = extgogit.ErrEmptyCommit
+	}
```
aryan9600, line 387: why are we creating an error here?

pjbgf: This is to be able to reuse the existing logic for handling empty commits.

Given that this controller will be refactored as part of the GA process, I found this was the easiest way to resolve the problem without duplicating code.

aryan9600 (approved):
lgtm! 🚀

hiddeco (approved):
Thank you @pjbgf :100:

stefanprodan (approved):

PR #473, opened 2022-12-22
build: Bump sigs.k8s.io/kustomize/kyaml

pjbgf:


stefanprodan:
We can’t bump kyaml, the kustomize/api needs to use it first

PR #485, opened 2023-02-13
Update image-reflector API to v1beta2

darkowlzz:


hiddeco (commented):
Wouldn't this require an update of the image-reflector image, etc. as well?

hiddeco (approved):
Never mind, it does not do something like this: https://github.com/fluxcd/helm-controller/blob/main/config/default/kustomization.yaml#L5-L6

PR #518, opened 2023-05-24
 Update controller-runtime, Kubernetes and kyaml dependencies

hiddeco:
This deals with various breaking changes in controller-runtime, as
documented in the release notes:
https://github.com/kubernetes-sigs/controller-runtime/releases/tag/v0.15.0

In short:

- `Watches` now use a `client.Object` instead of a `source.Kind`.
- `handler.MapFunc` signature accepts a Go context, which is used to
  log any errors, instead of silently ignoring them and/or panicking.
- Fake clients used in tests are now configured using
  `WithStatusSubresource` to enable the correct behavior for status
  updates and patches.
- Max concurrent reconciles and panic recovery is configured on the
  manager, instead of configuring them per reconciler instance.
- Various manager configuration options have been moved to new
  structures and/or fields.

In addition to this, all other dependencies which had updates are
updated to their latest (compatible) versions as well. Including an
update of kyaml to match Kustomize v5.

Part of: https://github.com/fluxcd/flux2/issues/3907 and https://github.com/fluxcd/flux2/issues/3564


stefanprodan:
Can you please mention kyaml as this is the major update, controller-runtime has no affect on end-users.

stefanprodan (approved):
LGTM

Thanks @hiddeco 🏅

PR #523, opened 2023-05-26
Update source-controller to v1.0.0-rc.4

hiddeco:


stefanprodan:
You could also bump IRC API

stefanprodan (approved):

PR #545, opened 2023-06-26
Align `go.mod` version with Kubernetes (Go 1.20)

stefanprodan:
Part of: https://github.com/fluxcd/flux2/issues/4014


makkes (approved):
Might want to change the [1.18 reference](https://github.com/fluxcd/image-automation-controller/blob/main/DEVELOPMENT.md#L27) in the dev docs as well.

Issue #491, opened 2023-03-05
Cryptic error message when there is no write access to SSH key


gecube:
Hi! 

I am creating this ticket as reminder and as a follow-up to the next discussion in Slack.

The original behaviour:

```
optimus  	imageupdateautomation/image-policy-settings	2023-02-27T23:11:59+01:00	False    	False	unknown error: remote: 	
querytool	imageupdateautomation/image-policy-settings	2023-02-27T23:21:22+01:00	False    	True 	no updates made   
```

```bash
$ flux version -o json
{
  "flux": "v0.38.2",
  "helm-controller": "v0.28.1",
  "image-automation-controller": "v0.28.0",
  "image-reflector-controller": "v0.23.1",
  "kustomize-controller": "v0.32.0",
  "notification-controller": "v0.30.2",
  "source-controller": "v0.33.0"
}
```

Image automation writes the cryptic error message "unknown error: remote:". It does not contain any clue about of the reason of the issue. In fact, I figured out that I forgot to add "write" permission to ssh key used to access the repo.

![image](https://user-images.githubusercontent.com/2912732/222979359-a95a5325-e9af-42e5-a8c7-7f5c2da4f097.png)

After I ticked the check box, the error message went away and the synchronisation of the image name to repo began to work.

### Expected behaviour:

Change the error message to something meaningful like "The ssh key does not have write permission. Please check the permissions of the key"

kingdonb:
Duplicate of:

* #208 
* #101 
* #102 

This should not be closed before we capture the error class somehow 👍 thank you for reporting

nick4fake:
Just spent about 2 hours with literally the same problem. Looks like message is trimmed.

As a temporary solution there might be some docs message that states "please check push acces, you might have forgotten it"

RogerSik:
Same error message
```
Events:
  Type     Reason  Age              From                         Message
  ----     ------  ----             ----                         -------
  Warning  error   2s (x2 over 5s)  image-automation-controller  unknown error: remote:
```
but my token has access to write to this repository and i have no clue where to looks for more messages.

mkarthiatgithub:
spend about 3 hours ...

PR #549, opened 2023-06-29
Update Git packages

aryan9600:
Update `fluxcd/pkg/git` and `fluxcd/pkg/git/gogit` which includes a QOL improvement for GitLab users accidentally using read-only deploy keys.

Fixes: https://github.com/fluxcd/image-automation-controller/issues/491


stefanprodan (approved):
LGTM

PR #550, opened 2023-06-30
Add backport GitHub Action workflow

hiddeco:
Part of: https://github.com/fluxcd/flux2/issues/4022


.github/labels.yaml
```diff
@@ -0,0 +1,11 @@
+# Configuration file to declaratively configure labels
+# Ref: https://github.com/EndBug/label-sync#Config-files
+
+- name: area/git
```
stefanprodan, line 4: Let's add here area/kyaml YAML patching released issues and PRs

stefanprodan (approved):

PR #573, opened 2023-08-15
Fix links in API docs

kingdonb:
* https://github.com/fluxcd/website/pull/1621

Unsure if maybe we should leave the `https://fluxcd.io` base domain intact here, so these docs are usable from the repo as a reference too. (Edit: I did wind up putting the base URL back in. I don't know if we're doing that consistently, but no need to break it here.)


kingdonb:
This one is OK to merge straightaway 👍

PR #581, opened 2023-09-04
Fix push branch reporting when is equal to checkout branch

### LINKED ISSUE #580: Branch name missing from commit event
https://github.com/fluxcd/image-automation-controller/commit/68ad71b1b3456d081e03bd027fbf7abb1b8ce0ed broke our tooling in 2 ways:
* The first line of the commit message changed format
* The branch mentioned in the commit message is now empty, for us

Ideally commit hash and branch name (and possibly other information as well) would be available as variables to use in the template, so we have full control of their output and we always know where to find it.

aryan9600:
Fix the push branch reported in the logs and status if `.spec.git.push.branch==.spec.git.checkout.branch`.

Fixes #580


stefanprodan (commented):
Can you please add a tests for the status message to verify that  it contains the new branch name and/or refspec.

stefanprodan (approved):
LGTM

Thanks @aryan9600 🏅

PR #604, opened 2023-11-20
Remove dependence on kustomize/setters2

squaremo:
The package setters2 has been removed from kustomize: https://github.com/kubernetes-sigs/kustomize/pull/5291

This commit removes the need to import setters2, by reproducing the last link with that code, which is a minor parsing helper.

I have not changed the comment explaining what was changed from the original, since it's still accurate. The parsing func is a buried detail.

### DIFF
--- a/pkg/update/filter.go
+++ b/pkg/update/filter.go
@@ -17,11 +17,12 @@ limitations under the License.
 package update
 
 import (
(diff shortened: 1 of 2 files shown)


stefanprodan (approved):
Thanks @squaremo

PR #633, opened 2024-02-01
Adding tests for accept function in the filter package

zhaque44:
**Tests for Accept - Scalar Node:**

Description: Tests the accept function when provided with a scalar node and valid setter schema.
The function should not return an error (assert.NoError).
The expected error value should be false (assert.Equal).

**Tests for Accept - Scalar Node - Error:**

Description: Tests the accept function when provided with a scalar node, with invalid setter schema.
The function should not return an error (assert.NoError).
The expected error value should be true (assert.Equal).


go.mod
```diff
@@ -33,6 +33,8 @@ require (
 	github.com/onsi/gomega v1.31.1
 	github.com/otiai10/copy v1.14.0
 	github.com/spf13/pflag v1.0.5
+	github.com/stretchr/testify v1.8.4
+	github.com/tj/assert v0.0.3
```
stefanprodan, line 37: Please remove these deps and use gomega, this the test framework we use for Flux.

zhaque44: thank you for pointing that out @stefanprodan

stefanprodan:
Please rebase with upstream main and force push, this PR should have a single commit before we can merge it. Thanks

PR #636, opened 2024-02-02
adding tests for accept function

zhaque44:
**Tests for Accept - Scalar Node:**

Description: Tests the accept function when provided with a scalar node and valid setter schema.
The function should not return an error (assert.NoError).
The expected error value should be false (assert.Equal).

**Tests for Accept - Scalar Node - Error:**

Description: Tests the accept function when provided with a scalar node, with invalid setter schema.
The function should not return an error (assert.NoError).
The expected error value should be true (assert.Equal).

### DIFF
--- a/pkg/update/filter_test.go
+++ b/pkg/update/filter_test.go
@@ -1,0 +1,55 @@
+package update
+
+import (
+	"testing"
+
+	"github.com/go-logr/logr"
+	. "github.com/onsi/gomega"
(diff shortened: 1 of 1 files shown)


zhaque44:
@stefanprodan I opened a new PR, disregard the last one

stefanprodan (approved):
LGTM

Thanks @zhaque44 🏅

Issue #437, opened 2022-09-01
Allow including the previous image tag in the commit message template


darkowlzz:
### Discussed in https://github.com/fluxcd/flux2/discussions/3027

<div type='discussions-op-text'>

<sup>Originally posted by **tun0** August 24, 2022</sup>
The notifications we currently have with Flux v1 (with help of fluxcloud) mention both old and new image tags.

With Flux v2, only the new image tag is mentioned.

It would be very nice if the previous tag would be made available as well in the notification object.</div>

darkowlzz:
Based on the discussion in https://github.com/fluxcd/image-reflector-controller/pull/311#issuecomment-1246683794 , this is not related to the image selected by the ImagePolicy, but it's about the old value being replaced by the new value in git manifest file.
An ImageUpdateAutomation notification can include the information about that old and new value in the manifest that it updated.

PR #642, opened 2024-03-04
Introduce ResultV2 for update results

darkowlzz:
`ResultV2` update result contains `Result` (original), which provides information in terms of files, objects and images, and also includes file changes which provides information in terms of files, objects and changes. The changes contain the old value, new value and the setter that was involved in the update.
`ResultV2` can be used to obtain detailed information about updates in terms of the previous value and the new value, which could be beneficial in commit messages to summarize the granular changes.

`ResultV2` is introduced in a backwards compatible way. The existing behavior is not changed. It will only be used by the refactored controller in the future. Introducing it in the main branch to keep the refactored controller changes small.

Related to https://github.com/fluxcd/image-automation-controller/issues/437

### DIFF
--- a/pkg/update/result.go
+++ b/pkg/update/result.go
@@ -96,3 +96,66 @@ func (r Result) Objects() map[ObjectIdentifier][]ImageRef {
 	}
 	return result
 }
+
+// ResultV2 contains Result of update and also the file changes made during the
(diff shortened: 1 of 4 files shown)


stefanprodan (approved):
LGTM

Great job on making this backwards compatible! Thanks @darkowlzz 🎖️

PR #682, opened 2024-05-12
Added table-driven tests with updated scenarios for ResultV2

zhaque44:
The ResultV2 struct has the following  **(AddChange, Changes, Objects)** in the update pkg do not have test coverage.

**TestResultV2_AddChange:** Tests the AddChange method to ensure that changes are correctly added to the result. It creates a new ResultV2, adds a change using AddChange method, and checks if the change was added correctly.

**TestResultV2_Changes:** Tests the Changes method to ensure that changes are returned correctly. It sets the FileChanges field of the ResultV2 struct, calls the Changes method, and verifies if the changes are returned.

**TestResultV2_Objects:** Tests the Objects method to ensure that objects are returned correctly. It sets the FileChanges field of the ResultV2 struct, calls the Objects method, and verifies if the objects are returned.

### DIFF
--- a/pkg/update/result_test.go
+++ b/pkg/update/result_test.go
@@ -98,16 +98,12 @@ func TestResultV2(t *testing.T) {
 
 	var result ResultV2
 	objectNames := []ObjectIdentifier{
-		{
-			yaml.ResourceIdentifier{
-				NameMeta: yaml.NameMeta{Namespace: "ns", Name: "foo"},
-			},
-		},
-		{
-			yaml.ResourceIdentifier{
-				NameMeta: yaml.NameMeta{Namespace: "ns", Name: "bar"},
-			},
-		},
+		{yaml.ResourceIdentifier{
+			NameMeta: yaml.NameMeta{Namespace: "ns", Name: "foo"},
+		}},
+		{yaml.ResourceIdentifier{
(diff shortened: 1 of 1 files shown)


zhaque44:
@stefanprodan can you please take a look when you have a minute

souleb:
thanks for the PR @zhaque44 but I don't really see why we should add it. Those functions are already tested in the existing test `TestResultV2`.

zhaque44:
> thanks for the PR @zhaque44 but I don't really see why we should add it. Those functions are already tested in the existing test `TestResultV2`.

Hi @souleb hope you are well very nice to meet you the original test is missing:

**Testing AddChange Method**
**Testing the Objects Method** <- these are not in the original test

Also, the test I have written focuses on the write assertions: Each subtest in the table driven test focuses on ensuring different aspects of ResultV2 are covered.

zhaque44:
@stefanprodan @souleb can we get a final review please, the assertions I’m adding bring value, if the team doesn’t think so I’ll close it. Please review the original test and what I have added. Thanks

souleb:
the only difference I see is that you add an assertion for `AddChange` where the original code does not. We could add that.

Other than that, the test cases presented are the same.

zhaque44:
@souleb my branch is in a bad state, will open a new PR with the changes:

PR #920, opened 2025-06-10
SparseCheckout when Spec.Update.Path specified

kane8n:
# Summary
If Spec.Update.Path is specified, only the manifest under this Path is required, no other files.
Therefore, SparseCheckout is valid.

Set the Path specified by Spec.Update.Path to SparseCheckoutDirectory so that only the necessary files are downloaded.


makkes (commented):
Thanks for this improvement. I would like to see this this feature aligned with the API for GitRepository resources that we implemented in https://github.com/fluxcd/source-controller/pull/1774 and make it an explicit field in the API instead of a feature flag of the controller. Not sure how @stefanprodan or @matheuscscp think about this.

stefanprodan:
> make it an explicit field in the API instead of a feature flag

We don't have a `path` field in GitRepository so we had to add a dedicated field for it. For image automation we know the path that we should set as sparse checkout as it's already in the API spec. Ideally image automation should always use sparse checkout but given that this feature is [experimental](https://git-scm.com/docs/git-sparse-checkout) in Git itself and brand new in go-git, I would not enable the feature flag by default. Sparse checkout can break at any time, the Git protocol may change, then the go-git project needs to adapt to it, etc.

internal/features/features.go
```diff
@@ -55,6 +58,10 @@ var features = map[string]bool{
 	// opt-out from v0.28
 	GitAllBranchReferences: true,
 
+	// GitSparseCheckout
+	// opt-out from v0.42
+	GitSparseCheckout: true,
```
stefanprodan, line 63: Let's make this opt-in for now. We can switch this to opt-out once sparse checkout is no longer an experimental feature in Git.

api/v1beta2/imageupdateautomation_types.go
```diff
@@ -90,6 +90,8 @@ type UpdateStrategy struct {
 	// Path to the directory containing the manifests to be updated.
 	// Defaults to 'None', which translates to the root path
 	// of the GitRepositoryRef.
+	// When a path is specified, sparse checkout will be enabled to limit the checkout scope
+	// to the specified directory, reducing clone size and time.
```
stefanprodan, line 94: I would remove this, since we'll have sparse checkout as an opt-in flag.

stefanprodan:
@kane8n please rebase

internal/source/source.go
```diff
@@ -196,6 +197,19 @@ func WithCheckoutOptionShallowClone() CheckoutOption {
 	}
 }
 
+// WithCheckoutOptionSparseCheckoutDirectories is a CheckoutOption option to configure
+// SparseCheckoutDirectories.
+func WithCheckoutOptionSparseCheckoutDirectories(updatePath string) CheckoutOption {
+	return func(cc *repository.CloneConfig) {
+		cleanedPath := filepath.Clean(updatePath)
+		if cleanedPath == "." {
```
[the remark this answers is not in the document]
kane8n: @stefanprodan 
Use filepath.Clean to remove the leading `./` prefix.
Should I use the same implementation as source-controller for consistency?

sample
- https://go.dev/play/p/syB7j09fMii

stefanprodan (approved):
LGTM

Thanks @kane8n 🏅 

PS. Please open a PR on the website repo to add feature gate to docs here: https://github.com/fluxcd/website/blob/main/content/en/flux/components/image/options.md#feature-gates

PR #942, opened 2025-07-16
Make `.spec.update.strategy` optional

stefanprodan:
Fix: #941


internal/controller/imageupdateautomation_controller_test.go
```diff
@@ -1468,6 +1468,70 @@ func TestImageUpdateAutomationReconciler_defaulting(t *testing.T) {
...
+		},
+	}
+	g.Expect(testEnv.Create(ctx, auto)).To(Succeed())
+	defer func() {
+		g.Expect(testEnv.Delete(ctx, auto)).To(Succeed())
+	}()
+
+	// Should default .spec.update to {strategy: Setters}.
+	var fetchedAuto imagev1.ImageUpdateAutomation
+	g.Eventually(func() bool {
```
matheuscscp, line 1524: nit: I think this doesn't need to be eventually as defaults are set synchronously during admission

matheuscscp (approved):
LGTM

internal/controller/imageupdateautomation_controller_test.go
stefanprodan: This is not about defaults, the `Get` goes via cache and we had issues before with testenv where `Get` would fail right after a `Create` due to Go routine contention in GHA runners.

PR #1035, opened 2026-05-29
Support signing commits with SSH key

hiddeco:
Depends on https://github.com/fluxcd/pkg/pull/1141 and https://github.com/fluxcd/pkg/pull/1222

### DIFF
--- a/api/v1/git.go
+++ b/api/v1/git.go
@@ -60,7 +60,8 @@ type CommitSpec struct {
 	// author of commits.
 	// +required
 	Author CommitUser `json:"author"`
-	// SigningKey provides the option to sign commits with a GPG key
+	// SigningKey provides the option to sign commits with an OpenPGP or
+	// SSH signing key, referenced from a Secret. See SigningKey.
 	// +optional
 	SigningKey *SigningKey `json:"signingKey,omitempty"`
 	// MessageTemplate provides a template for the commit message,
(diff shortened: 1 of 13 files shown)


api/v1/git.go
```diff
@@ -84,16 +85,49 @@ type CommitUser struct {
...
+	//
+	// Supported SSH key algorithms: ed25519, ecdsa-sha2-nistp256/384/521,
+	// and rsa (>= 2048-bit).
 	// +required
-	SecretRef meta.LocalObjectReference `json:"secretRef,omitempty"`
+	SecretRef meta.LocalObjectReference `json:"secretRef"`
+
+	// Type selects the signing-key format expected in the referenced
+	// Secret. Defaults to 'gpg'.
+	// +kubebuilder:default=gpg
```
matheuscscp: We don't normally set default values for API fields to avoid issues with SSA

api/v1beta2/git.go
matheuscscp, line 1: We don't normally add new features in previous API versions

matheuscscp (commented):
Mostly LGTM, a few important nits in the API package

api/v1beta2/git.go
stefanprodan: @hiddeco please undo changes in v1beta2, this API version is deprecated and will be dropped.

matheuscscp (approved):
LGTM! 🚀

internal/testutil/util.go
```diff
@@ -445,19 +450,90 @@ func GetSigningKeyPair(g *WithT, passphrase string) (*openpgp.Entity, []byte) {
...
+	gosshSigner, err := gossh.NewSignerFromKey(priv)
+	g.Expect(err).ToNot(HaveOccurred())
+
+	return pem.EncodeToMemory(pemBlock), gosshSigner.PublicKey()
+}
+
+// GetSSHSigningKeySecret returns a corev1.Secret in the shape the
+// controller expects for SSH signing: 'identity' holds the PEM-encoded
+// private key, 'password' holds the passphrase.
+func GetSSHSigningKeySecret(g *WithT, name, namespace string) (*corev1.Secret, gossh.PublicKey) {
```
stefanprodan, line 522: This looks like dead code, either remove it or refactor the tests to use it. 

PS. This helper sets a passphrase, but we don't have tests for encrypted keys in `imageupdateautomation_controller_test.go`, maybe we should.

internal/source/signing.go
```diff
@@ -0,0 +1,143 @@
...
+	"github.com/fluxcd/pkg/git/signature"
+
+	imagev1 "github.com/fluxcd/image-automation-controller/api/v1"
+)
+
+// Secret-data keys read by the signing pipeline.
+const (
+	signingSecretKeyGPG        = "git.asc"
+	signingSecretPassphraseGPG = "passphrase"
+	signingSecretKeySSH        = "identity"
```
matheuscscp, line 38: In the SC PR we used the extension `.sshpub` since GPG had an extension as well.

I think we should repeat the same here:

```suggestion
	signingSecretKeySSH        = "identity.sshpriv"
```

stefanprodan: I don't think we should be doing this here, the Flux standard for SSH key secrets is `identity`, `passphrase` and `known_hosts`.

matheuscscp: I this case we should not introduce `password` right below and should reuse `passphrase`

stefanprodan: Sorry, it's actually `password` in our SSH secrets.

hiddeco: The way this was designed to work is to allow pull SSH keys to be re-used as signing SSH keys, and to make the APIs of SC and IAC match.

stefanprodan (approved):
LGTM

Thanks @hiddeco 🥇

PR #1036, opened 2026-05-30
fix: propagate non-not-found errors when fetching GitRepository

immanuwell:
In `buildGitConfig`, non-not-found errors from `c.Get` (RBAC issue, transient API hiccup, etc.) are silently swallowed. The function then continues with an empty `GitRepository{}` and fails later with `failed to configure authentication options: no transport type set` instead of the real cause.

Root cause: `client.IgnoreNotFound(err) == nil` is true when the error *is* not-found, so the inner `return` only fires for that case. Non-not-found errors fall through with no return.

Fix adds the missing `return` for all other errors.

To reproduce: use `WithInterceptorFuncs` to inject a non-not-found error on `c.Get` for a `GitRepository`. The new test case `transient error fetching gitRepo` does exactly that and fails on unpatched code.

Also removes a stale `// var checkoutRef` comment.

## Test plan

- [x] `go test ./internal/source/... -run Test_buildGitConfig` passes
- [x] New test fails on unpatched code, passes with fix

### DIFF
--- a/internal/source/git.go
+++ b/internal/source/git.go
@@ -76,6 +76,7 @@ func buildGitConfig(ctx context.Context, c client.Client, originKey, srcKey type
 		if client.IgnoreNotFound(err) == nil {
 			return nil, fmt.Errorf("referenced git repository does not exist: %w", err)
 		}
+		return nil, fmt.Errorf("failed to get referenced git repository: %w", err)
 	}
 	cfg.url = repo.Spec.URL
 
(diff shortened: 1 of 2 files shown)


makkes:
Looks like this is a regression introduced in #647 where the controller had handled this properly:

```go
	if err := r.Get(ctx, originName, &origin); err != nil {
		if client.IgnoreNotFound(err) == nil {
			imagev1.SetImageUpdateAutomationReadiness(&auto, metav1.ConditionFalse, imagev1.GitNotAvailableReason, "referenced git repository is missing")
			log.Error(err, fmt.Sprintf("referenced git repository %s does not exist.", originName.String()))
			if err := r.patchStatus(ctx, req, auto.Status); err != nil {
				return ctrl.Result{Requeue: true}, err
			}
			return ctrl.Result{}, nil // and assume we'll hear about it when it arrives
		}
		return ctrl.Result{}, err
	}
```

makkes (approved):

makkes:
Thanks for your first contribution @immanuwell! 🙌🏻

PR #1038, opened 2026-06-04
Migrate event recorder to fluxcd/pkg/runtime/events.Recorder

### LINKED ISSUE #5761: Implement Kubernetes Event API v1
Transition the Flux ecosystem from legacy `core/v1` events to the modern `events.k8s.io/v1` API. The following changes have to follow a backward compatibility approach where legacy fields  may be bridged to the new schema (e.g., keep `InvolvedObject` regardless of being renamed, add a new `RelatedObject` field) to ensure no breaking changes for users relying on current event logs.

**Action Items**
- [x] Add a new AnnotatedEventRecorder interface upstream in `client-go/tools/events` - kubernetes/kubernetes#138103
- [x] Controller-Runtime - kubernetes-sigs/controller-runtime#3509
- [x] Update `fluxcd/pkg` to use `events.k8s.io/v1` - fluxcd/pkg#1144
- [ ] Update all controllers using the changes done in `runtime/events` from `fluxcd/pkg`. Use new field `relatedObject` to include object's source metadata (as InvolvedObject does) and `Action` as mandatory field.
   - [ ] Mandatory controllers
      - [ ] Notification-Controller - fluxcd/notification-controller#1267
         - [ ]  Adapt the event receiver to handle the events.k8s.io/v1 schema.
         - [ ] Update [spec/v1beta3/events.md](https://github.com/fluxcd/notification-controller/blob/main/docs/spec/v1beta3/events.md) to document new fields.
      - [ ] Helm-Controller fluxcd/helm-controller#1510
      - [ ] Kustomize-Controller fluxcd/kustomize-controller#1621
      - [ ] Source-Controller fluxcd/source-controller#2011
   - [ ] Optional controllers
      - [ ] Source-Watcher fluxcd/source-watcher/pull/343
      - [ ] Image-Automation-Controller fluxcd/image-automation-controller#1038
      - [ ] Image-Reflector-Controller fluxcd/image-reflector-controller#889

**Reference to the new API spec and changes**
- [API Deprecation Guide](https://kubernetes.io/docs/reference/using-api/deprecation-guide/#event-v125) ( `events.k8s.io/v1beta1` deprecated since k8s 1.25) 
- [KEP-383](https://github.com/kubernetes/enhancements/blob/master/keps/sig-instrumentation/383-new-event-api-ga-graduation/README.md): New Event API GA Graduation

adri1197:
- Replace kuberecorder.EventRecorder with fluxcd/events.Recorder to support structured event metadata including source object and action type.
- Upgrade event API from event/v1beta1 to event/v1
- Add action parameter (Reconciled, Failed, Waiting, Applied, Deleted, Progressing) to all event calls for richer event semantics.
- Pass source reference through reconcile, apply, prune, checkHealth, and finalize methods.
- Update tests to use events.Recorder.

Closes https://github.com/fluxcd/flux2/issues/5761

### DIFF
--- a/main.go
+++ b/main.go
@@ -26,7 +26,6 @@ import (
 	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
 	clientgoscheme "k8s.io/client-go/kubernetes/scheme"
 	_ "k8s.io/client-go/plugin/pkg/client/auth/gcp"
-	"k8s.io/utils/pointer"
 	ctrl "sigs.k8s.io/controller-runtime"
 	ctrlcache "sigs.k8s.io/controller-runtime/pkg/cache"
 	ctrlclient "sigs.k8s.io/controller-runtime/pkg/client"
@@ -37,6 +36,7 @@ import (
 	reflectorv1 "github.com/fluxcd/image-reflector-controller/api/v1"
 	"github.com/fluxcd/pkg/auth"
 	cache "github.com/fluxcd/pkg/cache"
+	"github.com/fluxcd/pkg/git"
 	"github.com/fluxcd/pkg/runtime/acl"
 	"github.com/fluxcd/pkg/runtime/client"
 	helper "github.com/fluxcd/pkg/runtime/controller"
@@ -48,8 +48,6 @@ import (
 	"github.com/fluxcd/pkg/runtime/pprof"
 	"github.com/fluxcd/pkg/runtime/probes"
 	sourcev1 "github.com/fluxcd/source-controller/api/v1"
-
-	"github.com/fluxcd/pkg/git"
 
 	imagev1 "github.com/fluxcd/image-automation-controller/api/v1"
 	"github.com/fluxcd/image-automation-controller/internal/features"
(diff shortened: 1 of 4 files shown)


matheuscscp:
Looks like we can do this here in this controller because it does not use `AnnotatedEventf` right?

adri1197:
> Looks like we can do this here in this controller because it does not use `AnnotatedEventf` right?

Correct — this controller only uses Eventf, so the migration is a direct swap 😃.

main.go
```diff
@@ -196,7 +195,7 @@ func main() {
 			ExtraHandlers: pprof.GetHandlers(),
 		},
 		Controller: ctrlcfg.Controller{
-			RecoverPanic:            pointer.Bool(true),
+			RecoverPanic:            ptr.To(true),
```
matheuscscp:
```suggestion
			RecoverPanic:            new(true),
```

matheuscscp (approved):
LGTM! 🚀

matheuscscp:
@adri1197 CI is failing, PTAL :pray:

adri1197:
> @adri1197 CI is failing, PTAL 🙏

Unfortunately, we may have to wait till https://github.com/kubernetes-sigs/controller-runtime/pull/3509 and therefore the upstream k8s MR, in order to merge https://github.com/fluxcd/pkg/pull/1144, which includes `github.com/fluxcd/pkg/apis/event/v1` interface.

PR #1042, opened 2026-06-06
fix: Preserve all object changes

immanuwell:
`.Changed.Objects` was overwriting earlier entries when the same object id showed up under more than one file.

This is triggerable in real repos, especially Kustomize style layouts where a base and an overlay both carry the same object. In that case commit templates only show part of the diff, which is pretty sneaky.

Repro:
1. Build an `update.Result` with the same `ObjectIdentifier` in two files.
2. Render a template that ranges over `.Changed.Objects`.
3. Before this patch, only the last file changes show up.
4. After this patch, both changes show up.

Tests:
- add a regression test for `Result.Objects()`
- add a template level regression test for `.Changed.Objects`

### DIFF
--- a/internal/source/source_test.go
+++ b/internal/source/source_test.go
@@ -35,12 +35,14 @@ import (
 	. "github.com/onsi/gomega"
 	"github.com/otiai10/copy"
 	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/types"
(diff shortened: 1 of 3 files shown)


matheuscscp (approved):
> which is pretty sneaky

Pretty sneaky indeed! Nice one!

LGTM! 🚀

